PR #27, opened 2024-01-22
Addresses

joneskm:


proto-types/build.rs
```diff
@@ -1,8 +1,8 @@
 use std::{env, error::Error};
 fn main() -> Result<(), Box<dyn Error>> {
-    let account_address_prefix = env::var("ACCOUNT_ADDRESS_PREFIX").map_err(|_| "ACCOUNT_ADDRESS_PREFIX environment variable must be set. This is best done in a .cargo/config.toml file in the root of your project")?;
+    let account_address_prefix = env::var("BECH_32_MAIN_PREFIX").map_err(|_| "BECH_32_MAIN_PREFIX environment variable must be set. This is best done in a .cargo/config.toml file in the root of your project")?;
     println!(
```
NYBACHOK, line 4: Is final application should have this `println!` too? If not, you probably need to use `dbg!`

proto-types/src/address.rs
```diff
@@ -8,23 +8,32 @@ use serde::{Deserialize, Deserializer, Serialize};
...
+const BECH_32_PREFIX_VAL_ADDR: &str = concat!(env!("BECH_32_MAIN_PREFIX"), "valoper");
 
-const ACCOUNT_ADDRESS_PREFIX: &str = env!("ACCOUNT_ADDRESS_PREFIX");
 const MAX_ADDR_LEN: u8 = 255;
 
-impl AccAddress {
-    pub fn into_inner(self) -> Vec<u8> {
-        self.0
-    }
+pub type AccAddress = BaseAddress<0>;
```
NYBACHOK, line 16: Newtype pattern instead of simple type alias?

proto-types/build.rs
joneskm: If those env variables are not set then the error message is displayed when building an application. Is that what you're concerned about?

proto-types/src/address.rs
joneskm: What benefits do we get from using the Newtype pattern?

joneskm: We would need to do the same for the ValAddress and for the other public methods of AccAddress.

Once more complex const parameter types arrive in Rust, we'll be able to replace `u8` with `&'static str`, see https://github.com/rust-lang/rust/issues/95174. This will allow for much cleaner code, something like:

```rust
pub type AccAddress = BaseAddress<BECH_32_PREFIX_ACC_ADDR>;
pub type ValAddress = BaseAddress<BECH_32_PREFIX_VAL_ADDR>;

#[derive(Debug, PartialEq, Clone)]
pub struct BaseAddress<const PREFIX: &'static str>(Vec<u8>);

impl<const PREFIX: &'static str> BaseAddress<PREFIX> {
    pub fn from_bech32(address: &str) -> Result<Self, AddressError> {
        let (hrp, data, variant) = bech32::decode(address)?;

        // This could be removed
        // let prefix = if PREFIX == 0 {
        //     BECH_32_PREFIX_ACC_ADDR
        // } else {
        //     BECH_32_PREFIX_VAL_ADDR
        // };


        if hrp != PREFIX {
            return Err(AddressError::InvalidPrefix {
                expected: PREFIX.into(),
                found: hrp,
            });
        };
```
I guess there's two questions:
1. Do we prefer the future const generic method using `str`or the newtype method?
2. In the meantime which method should we use?

NYBACHOK:
```rust
pub type AccAddress = BaseAddress<BECH_32_PREFIX_ACC_ADDR>;
pub type ValAddress = BaseAddress<BECH_32_PREFIX_VAL_ADDR>;
```
Looks much better.  I didn't think about it.

1. Depends on how quick this feature will arrive and MSV policy. 
2. Is it really important to use `str` over `[u8]`? If yes, then newtype

joneskm: IMO we should use const generics even though we are forced to use a `u8`

PR #31, opened 2024-01-30
[DRAFT] Remove for AVL Tree

NYBACHOK:


trees/src/iavl/tree.rs
```diff
@@ -14,7 +14,7 @@ use crate::{error::Error, merkle::EMPTY_HASH};
 use super::node_db::NodeDB;
 
 #[derive(Debug, Clone, PartialEq, Default)]
-pub(crate) struct InnerNode {
+pub struct InnerNode {
```
joneskm, line 17: Does this need to be public outside the crate?

trees/src/iavl/tree.rs
```diff
@@ -105,14 +122,37 @@ impl InnerNode {
 }
 
 #[derive(Debug, Clone, PartialEq, Default)]
-pub(crate) struct LeafNode {
+pub struct LeafNode {
```
joneskm, line 125: Same, does it need to be public?

trees/src/iavl/tree.rs
```diff
@@ -422,17 +567,141 @@ where
...
+                            }
+                        } {
+                            node.rebalance(version, node_db)?;
+                            return result;
+                        }
+                    }
+                },
+            }
+
+            // If control flow fell through to here, it's because we hit the Equal case above. The
```
joneskm, line 609: Can we avoid this assumption by moving the below code into the appropriate branch above?

trees/src/iavl/tree.rs
```diff
@@ -515,112 +784,46 @@ where
                 if balance_factor > 1 {
                     let left_node = root_node.get_mut_left_node(node_db);
 
-                    if &key < left_node.get_key() {
+                    if key[..] < *left_node.get_key() {
```
joneskm, line 787: Does this change now lead to a copy?

trees/src/iavl/tree.rs
```diff
@@ -669,7 +872,7 @@ impl<'a, T: RangeBounds<Vec<u8>>, R: Database> Range<'a, T, R> {
                                 .get_node(&inner.right_hash)
                                 .expect("node db should contain all nodes");
 
-                            self.delayed_nodes.push(right_node);
+                            self.delayed_nodes.push(*right_node);
```
joneskm, line 875: Should `delayed_node` be a vec of boxed `Nodes` to avoid this copy?

NYBACHOK: Changed to store boxes

trees/src/iavl/tree.rs
NYBACHOK: Returned to private

PR #33, opened 2024-02-07
Merge range for store

NYBACHOK:


store/src/store.rs
```diff
@@ -168,21 +168,29 @@ impl<DB: Database> KVStore<DB> {
...
-        //     MergedRange::merge(block_cached_values, persisted_values),
-        // );
-
-        self.persistent_store.range(range)
+        let tx_cached_values = self.tx_cache.range(range.clone());
+        let block_cached_values = self.block_cache.range(range.clone());
+        let mut persisted_values = self
+            .persistent_store
+            .range(range.clone())
+            .collect::<BTreeMap<_, _>>();
```
joneskm, line 180: This will pull all values from the persistent store (within the range) into memory. We want to avoid doing this.

PR #35, opened 2024-02-12
Tree remove

joneskm:


trees/src/iavl/tree.rs
```diff
@@ -1863,6 +1865,9 @@ mod tests {
                         .expect("node db should contain all nodes"),
                 };
 
+                println!("left_node: {:?}", left_node);
```
NYBACHOK, line 1868: For debug purposes, it's better to use[ `dbg!` ](https://doc.rust-lang.org/std/macro.dbg.html)macro

trees/src/iavl/node_db.rs
```diff
@@ -76,20 +76,20 @@ where
...
     }
 
-    pub(crate) fn get_node(&self, hash: &[u8; 32]) -> Option<Node> {
+    pub(crate) fn get_node(&self, hash: &[u8; 32]) -> Option<Box<Node>> {
         let cache = &mut self.cache.lock().expect("Lock will not be poisoned");
         let cache_node = cache.get(hash);
 
         if cache_node.is_some() {
-            return cache_node.map(|v| v.to_owned());
+            return cache_node.map(|v| Box::new(v.to_owned()));
```
NYBACHOK, line 82: Maybe it's better to cache boxed nodes?

trees/src/iavl/tree.rs
```diff
@@ -124,19 +171,174 @@ impl Default for Node {
...
+                    Self::left_rotate(left_node, version, node_db)
+                        .expect("given the imbalance, expect rotation to always succeed");
+                    Ok(Self::right_rotate(self, version, node_db)
+                        .expect("given the imbalance, expect rotation to always succeed"))
+                }
+                -1..=1 => {
+                    // The node is balanced
+                    Ok(())
+                }
+                _ => Err(Error::Balancing),
```
NYBACHOK, line 300: You could panic here, 'cause you already do this + this state should be unreachable.

NYBACHOK: Or instead of `.expect`, you could return error too

joneskm: It's only at the point the method is called can we be certain that we won't reach here. From within the method we cannot tell. This is different to the other panics in this method, where we know the rotation will succeed because we've checked (within this method) the tree's balance factor.

trees/src/iavl/node_db.rs
joneskm: I agree, they should probably be on the heap. I'll create a separate issue for this.

trees/src/iavl/tree.rs
NYBACHOK: Tree must be rotated when balance == -2/2. Otherwise, this is an issue

trees/src/iavl/node_db.rs
joneskm: Actually hash maps store their data on the heap so boxing doesn't change much in this regard. Was there some other reason for wanting to store boxed values?

NYBACHOK: You're right, but there is an issue with `Box`. When you do `Box::new(T)` T value is first allocated on stack then cloned onto heap.

joneskm: The get method of the cache returns a borrowed value, so we would still need a way to copy that borrowed value into a new box.

NYBACHOK: Okay. Let's leave it as is for now.

trees/src/iavl/tree.rs
NYBACHOK: For now, it's okay to leave this as is, but I think we could bring this topic later. Maybe even change interface of remove method for returning result without any panic inside

PR #36, opened 2024-02-12
Trait for different stores

NYBACHOK:


store/src/store.rs
```diff
@@ -236,7 +250,19 @@ impl<DB: Database> KVStore<DB> {
...
-pub(crate) enum AnyKVStore<'a, DB: Database> {
+/// Equivalent to [`BasicKVStore`](https://docs.cosmos.network/v0.46/core/store.html#base-layer-kvstores) from cosmos
+pub trait KVStoreTrait {
+    fn get(&self, k: &impl AsRef<[u8]>) -> Option<Vec<u8>>;
+    // fn contains( &self, k : &impl AsRef<[u8]>) -> bool;
+    fn set(&mut self, key: impl IntoIterator<Item = u8>, value: impl IntoIterator<Item = u8>);
+    // fn delete( &mut self, k : &impl AsRef<[u8]> );
+}
+
+pub trait AnyStoreReadTrait<DB: Database, SK> {
```
joneskm, line 261: I think we should get rid of this trait. It solves the same problem as the `KVStoreTrait` but in a way that's difficult to follow.

store/src/store.rs
```diff
@@ -236,7 +250,19 @@ impl<DB: Database> KVStore<DB> {
     }
 }
 
-pub(crate) enum AnyKVStore<'a, DB: Database> {
+/// Equivalent to [`BasicKVStore`](https://docs.cosmos.network/v0.46/core/store.html#base-layer-kvstores) from cosmos
+pub trait KVStoreTrait {
+    fn get(&self, k: &impl AsRef<[u8]>) -> Option<Vec<u8>>;
+    // fn contains( &self, k : &impl AsRef<[u8]>) -> bool;
+    fn set(&mut self, key: impl IntoIterator<Item = u8>, value: impl IntoIterator<Item = u8>);
```
joneskm, line 257: We can't have a set method since that will never work on a `QueryKVStore`
We should have a range method.

store/src/query_store.rs
```diff
@@ -36,11 +49,18 @@ pub struct QueryKVStore<'a, DB> {
...
-    pub fn get(&self, key: &[u8]) -> Option<Vec<u8>> {
-        self.persistent_store.get(key)
+impl<DB: Database> KVStoreTrait for QueryKVStore<'_, DB> {
+    fn get(&self, k: &impl AsRef<[u8]>) -> Option<Vec<u8>> {
+        self.persistent_store.get(k.as_ref())
     }
 
+    fn set(&mut self, _key: impl IntoIterator<Item = u8>, _value: impl IntoIterator<Item = u8>) {
+        // TODO
+        unimplemented!()
```
joneskm, line 59: See comment below. We can't have a set method on a `QueryKVStore`.

store/src/store.rs
NYBACHOK: Removed

store/src/store.rs
NYBACHOK: Before I add range method, we should merge [this](https://github.com/rumos-io/gears/pull/33) PR as it contains changes to range method

PR #44, opened 2024-02-28
Textual signing 2

joneskm:


gears/src/baseapp/baseapp.rs
```diff
@@ -119,7 +119,12 @@ impl<
 
         //TODO: handle request height > 1 as is done in SDK
 
-        let mut ctx = InitContext::new(&mut multi_store, self.get_block_height(), request.chain_id);
+        let chain_id = request.chain_id.try_into().unwrap_or_else(|_| {
+            error!("Invalid chain id provided by Tendermint.\nTerminating process\n");
```
NYBACHOK, line 123: What if we replace all `std::process::exit` with `expect` and use this crate for logs? 
https://docs.rs/log-panics/latest/log_panics/

x/auth/src/signing/renderer/value_renderer.rs
```diff
@@ -1,32 +1,49 @@
...
+    ) -> Result<Content, Error>;
 }
 
-/// The notion of "value renderer" is defined in ADR-050.
-pub trait ValueRenderer<SK: StoreKey, DB: Database> {
+#[cfg_attr(test, derive(PartialEq, Eq))]
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("not implemented")]
+    NotImplemented,
```
NYBACHOK, line 33: Why we need `NotImplemented` error?

x/auth/src/signing/renderer/primitives/bytes.rs
```diff
@@ -0,0 +1,101 @@
...
+                hasher.update(value);
+                let hashed = hasher.finalize();
+                let prefixed = format!("SHA-256={}", format_bytes(&hashed));
+                Ok(Content::new(prefixed).expect("prefixed is not empty"))
+            }
+        }
+    }
+}
+
+fn format_bytes(value: &[u8]) -> String {
```
NYBACHOK, line 30: As I remember this method made for `AccAdress` and used only with it. Do we need to separate them?

x/auth/src/signing/renderer/tx/tx.rs
```diff
@@ -0,0 +1,451 @@
...
+    renderer::value_renderer::{
+        DefaultPrimitiveRenderer, Error, PrimitiveValueRenderer, TryPrimitiveValueRenderer,
+        TryPrimitiveValueRendererWithMetadata, ValueRenderer,
+    },
+};
+
+/// Envelope is an internal data structure used to generate the tx envelope
+/// screens. Used in the same way as the Cosmos SDK Envelope type:
+/// https://github.com/cosmos/cosmos-sdk/blob/main/x/tx/signing/textual/tx.go
+pub struct Envelope<M> {
```
NYBACHOK, line 29: Interesting. I removed Envelope when worked on this

gears/src/baseapp/baseapp.rs
joneskm: We need the node to stop. At the moment if we panic it's only the thread that dies. Given that we use tendermint-abci I'm not sure how we would get round that?

x/auth/src/signing/renderer/value_renderer.rs
joneskm: Users must implement the ValueRenderer trait on their Messages. This gives them a way out.

x/auth/src/ante/ante.rs
```diff
@@ -60,7 +58,7 @@ pub trait AuthKeeper<SK: StoreKey>: Clone + Send + Sync + 'static {
 }
 
 pub trait AnteHandlerTrait<SK: StoreKey>: Clone + Send + Sync + 'static {
-    fn run<DB: Database, M: Message>(
+    fn run<DB: Database, M: Message + ValueRenderer>(
```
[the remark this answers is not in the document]
joneskm: We need them to implement ValueRendering so that we can call `sign_bytes_get` when the `sign_mode` is textual.

x/auth/src/signing/renderer/tx/tx.rs
joneskm: I think it's useful to have the envelope type to emphasize how transactions are structured for textual signing.

PR #62, opened 2024-04-02
Traits for context

NYBACHOK:


gears/src/types/context/mod.rs
```diff
@@ -1,5 +1,54 @@
...
+use proto_messages::cosmos::tx::v1beta1::tx_metadata::Metadata;
+use store_crate::{ReadKVStore, WriteKVStore};
+use tendermint::informal::{abci::Event, chain::Id};
+
 pub mod init_context;
 pub mod query_context;
-pub mod read_context;
 pub mod tx_context;
+
+pub trait Context<DB, SK> {
```
NYBACHOK, line 10: I'm think that `Context` name is wrong and it should be called QueryableContext. What do you think?

gears/src/types/context/mod.rs
```diff
@@ -1,5 +1,54 @@
...
+    fn height(&self) -> u64;
+    fn chain_id(&self) -> &Id;
+    fn metadata(&self) -> Metadata;
+}
+
+pub trait ContextMut<DB, SK>: Context<DB, SK> {
```
NYBACHOK, line 16: This should be renamed too, but I'm out of ideas

gears/src/types/context/mod.rs
```diff
@@ -1,5 +1,54 @@
...
+}
+
+pub trait ReadContext<DB: Database, SK> {
+    type KVStore: ReadKVStore<PrefixDB<DB>>;
+
+    ///  Fetches an immutable ref to a KVStore from the MultiStore.
+    fn kv_store(&self, store_key: &SK) -> &Self::KVStore; //AnyKVStore<'_, PrefixDB<DB>>;
+}
+
+pub trait WriteContext<DB: Database, SK>: ReadContext<DB, SK> {
```
NYBACHOK, line 28: This pair too doesn't explain much by their name

gears/src/types/context/mod.rs
NYBACHOK: Matter of taste. I use separate trait 'cause I don't want trait bounds to `Database & StoreKey` + as I remember, there should be cases where we only need access to KVStore.

store/src/types/any.rs
```diff
@@ -0,0 +1,44 @@
+use std::ops::RangeBounds;
+
+use database::Database;
+use trees::iavl::Range;
+
+use crate::ReadKVStore;
+
+use super::{kv::KVStore, query::kv::QueryKVStore};
+
+pub enum AnyKVStore<'a, DB> {
```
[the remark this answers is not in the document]
NYBACHOK: I left this enum only in prefix, so yes

gears/src/types/context/mod.rs
joneskm: But I don't think we'll ever have a type which implements the `ReadTrait` and doesn't implement the `Context` trait.

NYBACHOK: Unified this trait with `ReadContext`

gears/src/types/context/mod.rs
```diff
@@ -1,5 +1,50 @@
...
+pub trait Context<DB: Database, SK> {
+    type KVStore: ReadKVStore<DB>;
+
+    ///  Fetches an immutable ref to a KVStore from the MultiStore.
+    fn kv_store(&self, store_key: &SK) -> &Self::KVStore; //AnyKVStore<'_, PrefixDB<DB>>;
+
+    fn height(&self) -> u64;
+    fn chain_id(&self) -> &Id;
+    fn metadata(&self) -> Metadata;
+}
```
joneskm, line 19: I think it might make more sense to move, `height` and `chain_id` to the `ReadContext` trait and rename this trait to `WriteContext`. Also we can probably get rid of metadata completely.

NYBACHOK: But I removed `ReadContext` trait

PR #63, opened 2024-04-09
Refactor client transaction and query code

dmvict:
I've committed the `Cargo.lock` because the `.gitignore` doesn't ignore the file. 

The changes are around method `create_signed_transaction`, `handle_tx` and `handle_query`. I moved some code to types constructors. The names of constructors are very close to its purpose. It allows us to remove `TODO`'s for hardcoded values. 
Also, I created a lot of simpler methods and functions that can be reused anywhere because they are context independent. 

Let's discuss the PR.


gears/src/crypto.rs
```diff
@@ -25,47 +17,27 @@ pub struct SigningInfo {
...
     let body_bytes = tx_body.encode_vec();
     let auth_info_bytes = auth_info.encode_vec();
 
-    let mut sign_doc = SignDoc {
-        body_bytes: body_bytes.clone(),
-        auth_info_bytes: auth_info_bytes.clone(),
-        chain_id: chain_id.into(),
-        account_number: 0, // This gets overwritten
-    };
+    let mut sign_doc = SignDoc::new_with_opt_acc_number(
```
joneskm, line 28: This method doesn't exist

gears/src/application/handlers.rs
```diff
@@ -37,15 +46,7 @@ pub trait TxHandler {
         chain_id: tendermint::informal::chain::Id,
         fee: Option<SendCoins>,
     ) -> anyhow::Result<Response> {
-        let fee = Fee {
-            amount: fee,
-            gas_limit: 100000000, //TODO: remove hard coded gas limit
```
joneskm, line 42: Some of theses values need to be added as CLI params - that's what the TODO is alluding to.

gears/src/application/handlers.rs
```diff
@@ -54,24 +55,30 @@ pub trait TxHandler {
...
 
-        let client = HttpClient::new(tendermint::rpc::Url::try_from(node)?)?;
+        let raw_tx = create_signed_transaction(&[signing_info], auth_info, tx_body, chain_id);
 
-        broadcast_tx_commit(client, raw_tx)
+        commit_transaction(raw_tx, node)
     }
 }
 
+#[inline]
```
joneskm, line 76: Is it a good idea to be manually inlining functions? see https://matklad.github.io/2021/07/09/inline-in-rust.html

gears/src/application/handlers.rs
```diff
@@ -141,7 +158,7 @@ pub trait AuxHandler {
 }
 
 // TODO: we're assuming here that the app has an auth module which handles this query
-fn get_account_latest(address: AccAddress, node: &str) -> anyhow::Result<QueryAccountResponse> {
+pub fn get_account_latest(address: AccAddress, node: &str) -> anyhow::Result<QueryAccountResponse> {
```
joneskm, line 161: Does this need to be public?

gears/src/application/handlers.rs
```diff
@@ -54,24 +55,30 @@ pub trait TxHandler {
...
 
-        let tx_body = TxBody {
-            messages: vec![msg],
-            memo: String::new(),                    // TODO: remove hard coded
-            timeout_height: 0,                      // TODO: remove hard coded
-            extension_options: vec![],              // TODO: remove hard coded
-            non_critical_extension_options: vec![], // TODO: remove hard coded
-        };
-
+        let fee = Fee::new_with_defaults(fee);
```
joneskm, line 58: I'm not sure how valuable this method will be once we add more params to the CLI?

proto-messages/src/cosmos/tx/v1beta1/tx/tx_body.rs
```diff
@@ -38,6 +38,20 @@ pub struct TxBody<M: Message> {
     pub non_critical_extension_options: Vec<Any>, //TODO: use a domain type here
 }
 
+/// The implementation that is intended to create default tx body message from a single message.
+/// All fields except `messages` has default value.
+impl<M: Message> From<M> for TxBody<M> {
```
joneskm, line 43: Again, not sure how useful this will be once we add params to the CLI

joneskm (commented):
Thanks for the PR. I've left a few comments.

dmvict:
Hello @joneskm 

Thank you for the review. I agree that many things looks like extra moves. And inside the Gears it has more sense. 

I'll try to explain the logic. 
I proposed the changes because in the `ultradao` project I reimplemented a lot of client code. Most of the code in the `ultradao` is a copy of original logic with small changes. You can visit links 
https://github.com/aragonzkresearch/ultradao/blob/master/ultradao/src/client/transaction.rs
https://github.com/aragonzkresearch/ultradao/blob/master/gateway/src/query.rs
https://github.com/aragonzkresearch/ultradao/blob/master/gateway/src/handlers/transaction.rs
and find duplications. 

I've tried to extract reusable parts in the sdk and make it public. It can be reused in the app.

I'm calm if most of the changes will be rejected. Just let's save default entity constructors, please. The constructed entities can be modified in the code after initialization. 

I'll answer the comments. It's like a summary.

gears/src/application/handlers.rs
dmvict: You're right. 
It's easy way to construct default instance and reuse it anywhere. 

In Rust world, the builder pattern is widely used. I like it for such cases when we can setup field by field. But default constructors are much simpler.

gears/src/application/handlers.rs
dmvict: The rule:
> Fourth, when building libraries, proactively add #[inline] to small non-generic functions. Pay special attention to impls: Deref, AsRef and the like often benefit from inlining. A library can’t anticipate all usages upfront, it makes sense to not prematurely pessimize future users. Note that #[inline] is not transitive: if a trivial public function calls a trivial private function, you need to #[inline] both. See [this benchmark](https://github.com/matklad/benchmarks/tree/91171269f0a6e260a27111d07661021a89d20085/rust-inline) for details.

is applicable for the case. I'm force inline for small functions. But you're right, it can be skipped.

gears/src/application/handlers.rs
dmvict: I checked the code, I used it at least twice. 

Here I restored the original version. I'll use the alternative implementation in `ultradao`.

gears/src/application/handlers.rs
dmvict: It's a way to initialize instance and then setup it's fields. 

Sure, we can get the values from the CLI.

Please, add comment: Should I remove default constructors from `Fee` and `TxBody`? I'll implement changes in the next commit.

dmvict:
Closed due to changes in module and removing main changes.

PR #69, opened 2024-04-16
WIP: Implement gas metering

NYBACHOK:


gears/src/baseapp/mod.rs
```diff
@@ -42,6 +47,7 @@ pub struct BaseApp<
 > {
     multi_store: Arc<RwLock<MultiStore<RocksDB, SK>>>,
     height: Arc<RwLock<u64>>,
+    block_gas_meter: Arc<RwLock<Box<dyn GasMeter>>>,
```
joneskm, line 50: Does check_tx need a block_gas_meter? If it does it can't use the same one as deliver_tx.

x/auth/src/keeper.rs
```diff
@@ -217,35 +218,3 @@ fn create_auth_store_key(address: AccAddress) -> Vec<u8> {
...
-//     }
-
-//     pub fn get_permissions(&self) -> Vec<String> {
-//         match self {
-//             Module::FeeCollector => vec![],
-//         }
-//     }
-// }
-
-//TODO: copy tests across
```
joneskm, line 251: Should keep the TODO about copying tests.

joneskm: A long time ago the auth keeper had tests! https://github.com/rumos-io/gears/blob/3455cee74dde764948e87cd4590c8f885f7e9b92/gears/src/x/auth/auth.rs#L192
That's what `TODO: copy tests across` is referring to.

NYBACHOK:
Duplicate of https://github.com/rumos-io/gears/pull/73

PR #76, opened 2024-04-29
Merge range for store, but with Cow

NYBACHOK:


store/src/types/query/kv.rs
```diff
@@ -25,8 +26,13 @@ impl<'a, DB: Database> QueryableKVStore<DB> for QueryKVStore<'a, DB> {
...
     }
 
-    fn range<R: RangeBounds<Vec<u8>> + Clone>(&self, range: R) -> Range<'_, R, DB> {
-        self.persistent_store.range(range)
+    fn range<R: RangeBounds<Vec<u8>> + Clone>(&self, range: R) -> MergedRange<'_> {
+        MergedRange::merge(
+            vec![].into_iter(),
+            self.persistent_store
+                .range(range)
+                .map(|(second, first)| (Cow::Owned(first), Cow::Owned(second))),
```
joneskm, line 34: This is the wrong way round. We should copy the tests for the regular KV store to test QueryKVStore.

PR #81, opened 2024-05-09
chore: Add method `end_block` and refactor keeper module

dmvict:
Note that I added the casting between `PublicKey` types https://github.com/rumos-io/gears/compare/main...dmvict:staking?expand=1#diff-154e1f7fcd32f0f2dfb9243e9fe95f8d1d91c78bec9c6562e7bde9dc992395daR35


x/staking/src/abci_handler.rs
```diff
@@ -53,9 +53,11 @@ impl<
     }
 
     fn init_genesis<DB: Database>(&self, ctx: &mut InitContext<'_, DB, SK>, genesis: GenesisState) {
-        self.keeper
+        // TODO
+        let _updates = self
+            .keeper
             .init_genesis(ctx, genesis)
-            .expect("Cannot perform genesis actions.")
+            .expect("Cannot perform genesis actions.");
```
joneskm, line 60: When you use Result::expect(), could you explain in the message why you think it's safe to do so? https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

x/staking/src/proto.rs
```diff
@@ -45,7 +78,7 @@ pub struct Validator {
     /// unbonding_height defines, if unbonding, the height at which this validator has begun unbonding.
     pub unbonding_height: i64,
     /// unbonding_time defines, if unbonding, the min time for the validator to complete unbonding.
-    pub unbonding_time: std::time::Duration,
+    pub unbonding_time: chrono::DateTime<Utc>,
```
joneskm, line 81: `pub consensus_pubkey: PublicKey,` above, is there a reason why this isn't a Tendermint public key?

x/staking/src/keeper/validators_and_total_power.rs
```diff
@@ -0,0 +1,89 @@
...
+    > Keeper<SK, PSK, AK, BK, KH>
+{
+    /// Load the last total validator power.
+    pub fn get_last_total_power<DB: Database, CTX: QueryableContext<DB, SK>>(
+        &self,
+        ctx: &CTX,
+    ) -> Option<Uint256> {
+        let store = ctx.kv_store(&self.store_key);
+        store.get(&LAST_TOTAL_POWER_KEY).map(|bytes| {
+            Uint256::from_be_bytes(bytes.try_into().expect("Unexpected conversion error."))
```
joneskm, line 18: Same as above with use of expect. If this can only happen because there's either a bug in your code or the data is corrupt then we have this trait  `gears::store::database::ext::UnwrapCorrupt;` to captue that.

x/staking/src/keeper/traits.rs
```diff
@@ -0,0 +1,170 @@
...
+    ) -> Option<ModuleAccount>;
+
+    fn set_module_account<DB: Database, CTX: TransactionalContext<DB, SK>>(
+        &self,
+        context: &mut CTX,
+        acc: ModuleAccount,
+    );
+}
+
+/// BankKeeper defines the expected interface needed to retrieve account balances.
```
joneskm, line 28: In future we can move this into Gears... it's OK here for now.

x/staking/src/proto.rs
```diff
@@ -54,25 +87,88 @@ pub struct Validator {
...
-            .expect("Unexpected conversion error");
-        amount / 10i64.pow(6)
+            .expect("Unexpected conversion error")
+    }
+
+    /// GetValidatorsByPowerIndexKey creates the validator by power index.
+    /// Power index is the key used in the power-store, and represents the relative
+    /// power ranking of the validator.
+    /// VALUE: validator operator address ([]byte)
+    pub fn key_by_power_index_key(&self, power_reduction: i64) -> Vec<u8> {
```
joneskm, line 153: Nit: could we use an array instead of a vec here i.e. does this return a fixed length?

joneskm (commented):
Thanks Dmytro. Overall this looks good. I've left a few minor comments.

x/staking/src/abci_handler.rs
dmvict: @joneskm  Thank you for the feedback. 

I added detailed explanation for each `expect` and updated some handlers.  Please, check it. If you'll notice any unclear explanation, add a comment, please,

x/staking/src/proto.rs
dmvict: I had no particular reason to use the domain `PublicKey` instead of tendermint `PublicKey`. I suggested that it would be better to use domain types and used it. 

Should I change the type to tendermint `PublicKey` from https://github.com/rumos-io/gears/blob/main/tendermint/src/types/proto/crypto.rs#L2?

x/staking/src/keeper/validators_and_total_power.rs
dmvict: I added detailed explanation for each `expect` and updated some handlers.  Please, check it. If you'll notice any unclear explanation, add a comment, please,

x/staking/src/keeper/traits.rs
dmvict: The staking module in sdk owns this trait. 
We can move it or just use the from the module.

x/staking/src/proto.rs
dmvict: @joneskm We can use an array if ValAddress has constant length. As I understand the length of the address is not static but restricted by the constant https://github.com/rumos-io/gears/blob/main/tendermint/src/types/proto/crypto.rs#L2. 

Am I correct?

I updated the realization of method `get_unbonding_delegation_time_key` https://github.com/rumos-io/gears/pull/81/commits/9fa274485ac5de16040803ddcf7f8d4fe6b06b80#diff-d215ed340a59cbde946a9a09538396a743896c746fe271374eb0e8a71c2e3635R404

x/staking/src/keeper/bonded.rs
```diff
@@ -0,0 +1,85 @@
...
+impl<
+        SK: StoreKey,
+        PSK: ParamsSubspaceKey,
+        AK: AccountKeeper<SK>,
+        BK: BankKeeper<SK>,
+        KH: KeeperHooks<SK>,
+    > Keeper<SK, PSK, AK, BK, KH>
+{
+    /// get_bonded_pool returns the bonded tokens pool's module account
+    pub fn get_bonded_pool<DB: Database, CTX: TransactionalContext<DB, SK>>(
```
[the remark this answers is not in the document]
dmvict: @NYBACHOK, Thank you for the feedback. 

I promise to remove the prefix in the next PR. I want to have minimal merge conflicts.

x/staking/src/keeper/mod.rs
```diff
@@ -0,0 +1,524 @@
...
+mod traits;
+mod unbonded;
+mod unbonding;
+mod validator;
+mod validators_and_total_power;
+pub use traits::*;
+use unbonding::*;
+use validator::*;
+
+const POOL_KEY: [u8; 1] = [0];
```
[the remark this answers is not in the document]
dmvict: @NYBACHOK. I promise to move constants. I want to have minimal merge conflicts.

x/staking/src/proto.rs
joneskm: The application keys (the ones in Gears) should be separate to the consensus keys (tendermint). You're right we should generally use domain types. I suggest we create a domain type for the tendermint pub key in `gears::tendermint.`

x/staking/src/proto.rs
joneskm: Ah OK. The ValAddress is not a constant length so ignore my comment.

x/staking/src/abci_handler.rs
```diff
@@ -53,9 +53,16 @@ impl<
...
-            .init_genesis(ctx, genesis)
-            .expect("Cannot perform genesis actions.")
+        // TODO
+        if let Err(err) = self.keeper.init_genesis(ctx, genesis) {
+            panic!(
+                "It is important to perform genesis actions.
+            During genesis staking module initialize pools and queues
+            using the parameters provided by the application.
+            Original error: {:?}",
+                err
```
joneskm, line 63: You shouldn't need to panic here. The keeper `init_genesis` method should be infallible.

x/staking/src/abci_handler.rs
```diff
@@ -68,11 +75,19 @@ impl<
...
         _request: RequestEndBlock,
     ) -> Vec<ValidatorUpdate> {
-        vec![]
-        // self.keeper.end_block(ctx, request)
+        match self.keeper.block_validator_updates(ctx) {
+            Ok(updates) => updates,
+            Err(err) => panic!(
+                "Error thrown in method 'end_block'. It can broke blockchain.
+                Original error: {:?}",
+                err
```
joneskm, line 86: Similar to above, the keeper `block_validator_updates` method should be infallible.

x/staking/src/keeper/delegation.rs
```diff
@@ -0,0 +1,23 @@
...
+    pub fn set_delegation<DB: Database, CTX: TransactionalContext<DB, SK>>(
+        &self,
+        ctx: &mut CTX,
+        delegation: &Delegation,
+    ) -> anyhow::Result<()> {
+        let store = ctx.kv_store_mut(&self.store_key);
+        let mut delegations_store = store.prefix_store_mut(DELEGATIONS_KEY);
+        let mut key = delegation.delegator_address.to_string().as_bytes().to_vec();
+        key.put(delegation.validator_address.to_string().as_bytes());
+        delegations_store.set(key, serde_json::to_vec(&delegation)?);
```
joneskm, line 20: Do you think we should return an error here or unwrap through a call to `expect`? We know that the serialization of the Delegation struct doesn't fail.

x/staking/src/keeper/redelegation.rs
```diff
@@ -0,0 +1,189 @@
...
+        completion_time: chrono::DateTime<Utc>,
+        redelegations: Vec<DvvTriplet>,
+    ) -> anyhow::Result<()> {
+        let store = ctx.kv_store_mut(&self.store_key);
+        let mut store = store.prefix_store_mut(REDELEGATION_QUEUE_KEY);
+
+        let key = completion_time
+            .timestamp_nanos_opt()
+            .expect("The timestamp_nanos_opt produces an integer that represents time in nanoseconds.
+                    The error in this method means that some system failure happened and the system cannot continue work.")
```
joneskm, line 144: The expect message should be something like "provided the date is between 1677-09-21T00:12:43.145224192 and 2262-04-11T23:47:16.854775807 the conversion won't fail"

joneskm: I'm starting to doubt whether panicking is the right thing to do here. It could be a vulnerability that allows attackers to halt a chain. Where does the `completion_time` come from? Perhaps we should parse those times upfront to guarantee that they can be converted to an i64?

PR #84, opened 2024-05-14
Changes to support the staking module

joneskm:
Move Address types into gears - this allows the From trait to be implkemented on TM keys. Since QueryAccountRequest depended on AccAddress this type is also moved into gears.

Add ConsAddress type.

Cleanup TM key struct to be more usable.

Remove protobuf ser/deser from some types in TM since they aren't needed.


tendermint/src/types/proto/crypto.rs
```diff
@@ -1,21 +1,53 @@
...
+            },
         }
     }
 }
 
 impl From<inner::PublicKey> for PublicKey {
     fn from(inner::PublicKey { sum }: inner::PublicKey) -> Self {
-        Self {
-            sum: sum.map(Into::into),
+        let Some(sum) = sum else {
```
NYBACHOK, line 42: Well... I won't agree with you on this. This is unexpected behavior for `From` implementation. Better use method name `parse_or_terminate` or handle this logic on upper levels.

tendermint/src/types/request/init_chain.rs
```diff
@@ -3,19 +3,13 @@ use crate::types::{
     time::Timestamp,
 };
 
-#[derive(Clone, PartialEq, Eq, ::prost::Message, serde::Serialize, serde::Deserialize)]
```
NYBACHOK, line 6: Be aware that `prost::Message` implements `Debug` too

NYBACHOK (changes requested):

x/staking/src/proto.rs
```diff
@@ -68,11 +68,13 @@ impl Validator {
...
-            .expect("Unexpected conversion error");
-        amount / 10i64.pow(6)
+        // let amount = self
+        //     .tokens
+        //     .amount
+        //     .parse::<i64>()
+        //     .expect("Unexpected conversion error");
+        // amount / 10i64.pow(6)
+        //TODO: original code above doesn't compile
+        12
```
dmvict, line 78: It seems like your code doesn't compile here.

tendermint/src/types/proto/crypto.rs
joneskm: I agree... I've pushed a change

gears/src/types/address.rs
```diff
@@ -167,6 +184,44 @@ impl<const PREFIX: u8> From<BaseAddress<PREFIX>> for Vec<u8> {
     }
 }
 
+impl From<tendermint::types::proto::crypto::PublicKey> for ConsAddress {
```
dmvict, line 187: It is a question. 
Can we transform tendermint PublicKey into domain PublicKey? It allows us to generate any type of addresses.

tendermint/src/types/proto/crypto.rs
```diff
@@ -1,21 +1,53 @@
...
+    Ed25519(Vec<u8>), //TODO: should we check that bytes contain a valid public key?
+    #[serde(
+        rename = "tendermint/PubKeySecp256k1",
+        with = "crate::types::serializers::bytes::base64string"
+    )]
+    Secp256k1(Vec<u8>), //TODO: should we check that bytes contain a valid public key?
+}
+
+impl PublicKey {
+    pub fn raw(&self) -> Vec<u8> {
```
dmvict, line 19: I think we can return slice of bytes instead of full copy. 

```rs 
pub fn as_bytes(&self) -> &[u8] 
```

tendermint/src/types/proto/crypto.rs
```diff
@@ -37,19 +37,28 @@ impl From<PublicKey> for inner::PublicKey {
...
+    fn try_from(inner::PublicKey { sum }: inner::PublicKey) -> Result<Self, Self::Error> {
+        let sum = sum.ok_or(Error::InvalidData("public key is empty".to_string()))?;
         match sum {
-            inner::Sum::Ed25519(value) => PublicKey::Ed25519(value),
-            inner::Sum::Secp256k1(value) => PublicKey::Secp256k1(value),
+            inner::Sum::Ed25519(value) => Ok(PublicKey::Ed25519(value)),
+            inner::Sum::Secp256k1(value) => Ok(PublicKey::Secp256k1(value)),
         }
     }
+    // fn from(inner::PublicKey { sum }: inner::PublicKey) -> Self {
```
NYBACHOK, line 50: Don't forget to clear this comments

NYBACHOK (approved):

tendermint/src/types/proto/crypto.rs
joneskm: Agree

tendermint/src/types/request/init_chain.rs
joneskm: Yes, we can add Debug derive if/when necessary.

x/staking/src/proto.rs
joneskm: Yes, not sure why... we can fix it later

gears/src/types/address.rs
joneskm: Tendermint has it's own domain PublicKey separate to Cosmos. I think it's cleaner to keep them separate. I'm not sure generating all the addresses makes sense for the two types of key e.g. I don't think a ConsAddress (consensus address) should be generated for a Cosmos PublicKey.

dmvict (approved):

PR #90, opened 2024-05-17
WIP: Mempool gas validation

NYBACHOK:


gears/src/x/ante.rs
```diff
@@ -142,6 +146,56 @@ impl<AK: AuthKeeper<SK>, BK: BankKeeper<SK>, SK: StoreKey, GC: SignGasConsumer>
...
+                    amount: gp
+                        .amount
+                        .checked_mul_ceil(cosmwasm_std::Decimal::new(u64::from(gas).into()))?, // TODO:NOW Great. Replace it with something sane
+                });
+            }
+
+            if fee_coins.inner().into_iter().any(|fee_coin| {
+                required_fees
+                    .iter()
+                    .any(|req_fee| req_fee.amount >= fee_coin.amount)
```
joneskm, line 186: Couldn't these be different denominations? For every required fee we need to check that the fee contains at least that much of that denomination.

joneskm (commented):
Probably as part of a separate PR we should add a min gas price field in the config

gears/src/x/ante.rs
NYBACHOK: You're right, I found [where](https://github.com/cosmos/cosmos-sdk/blob/d3f09c222243bb3da3464969f0366330dcb977a8/types/coin.go#L572) it does it

gears/src/x/ante.rs
```diff
@@ -142,6 +146,56 @@ impl<AK: AuthKeeper<SK>, BK: BankKeeper<SK>, SK: StoreKey, GC: SignGasConsumer>
...
+        let fee = tx.auth_info.fee.amount.as_ref();
+        let gas = tx.auth_info.fee.gas_limit;
+
+        let min_gas_prices = ctx.options.min_gas_prices();
+
+        if min_gas_prices.is_empty() || min_gas_prices.iter().any(|this| this.amount.is_zero()) {
+            return Ok(());
+        }
+
+        if let Some(fee_coins) = fee {
```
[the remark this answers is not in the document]
NYBACHOK: I'm not sure about this. For now in skipped

gears/src/x/ante.rs
```diff
@@ -142,6 +146,56 @@ impl<AK: AuthKeeper<SK>, BK: BankKeeper<SK>, SK: StoreKey, GC: SignGasConsumer>
...
+            let mut required_fees = Vec::with_capacity(min_gas_prices.len());
+
+            for gp in min_gas_prices {
+                required_fees.push(Coin {
+                    denom: gp.denom,
+                    amount: gp
+                        .amount
+                        .checked_mul_ceil(cosmwasm_std::Decimal::new(u64::from(gas).into()))?, // TODO:NOW Great. Replace it with something sane
```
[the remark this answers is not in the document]
NYBACHOK: It implements `Fraction` trait... u64 doesn't

gears/src/x/ante.rs
joneskm: OK, so we must do the same

gears/src/x/ante.rs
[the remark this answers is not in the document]
joneskm: The logic should be: if the required gas fee > 0 and the fee is None the tx is rejected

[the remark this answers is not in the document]
NYBACHOK: Okay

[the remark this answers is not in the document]
NYBACHOK: Corrected

gears/src/x/ante.rs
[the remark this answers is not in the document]
NYBACHOK: Corrected those too

gears/src/x/ante.rs
```diff
@@ -142,6 +147,70 @@ impl<AK: AuthKeeper<SK>, BK: BankKeeper<SK>, SK: StoreKey, GC: SignGasConsumer>
...
+
+            if !is_any_gte(fee_coins.inner(), &required_fees) {
+                Err(anyhow!(
+                    "insufficient fees; got: {:?} required: {:?}",
+                    fee_coins,
+                    required_fees
+                ))?
+            }
+        } else {
+            Err(anyhow!("rejected. `fee_coins` is None"))?
```
joneskm, line 193: This isn't quite right. It's OK for the fee to be `None` provided the required fees are all zero.

gears/src/types/base/coin.rs
```diff
@@ -18,6 +18,22 @@ pub struct Coin {
     pub amount: Uint256,
 }
 
+impl Coin {
+    pub fn amount_of<'a, I: Iterator<Item = &'a Coin>>(coins: I, denom: &Denom) -> Uint256 {
```
joneskm, line 22: What happens if the same denom is repeated? As above I think we should have a new type which captures the fact that there are no repeated denoms in a list. This function should be a method on that type.
There is no distinction between denom not found and finding the denom with an amount zero. This might be OK for how you use it but will likely cause problems in future. Better to be explicit and return an option?

gears/src/baseapp/options.rs
```diff
@@ -0,0 +1,25 @@
...
+struct InnerOptions {
+    pub min_gas_prices: Vec<DecimalCoin>,
+}
+
+impl NodeOptions {
+    pub fn new(min_gas_prices: Vec<DecimalCoin>) -> Self {
+        Self(Arc::new(RwLock::new(InnerOptions { min_gas_prices })))
+    }
+
+    pub fn min_gas_prices(&self) -> Vec<DecimalCoin> {
```
joneskm, line 18: We probably want to avoid repeated coins (more than one coin with the same denom). Better to capture this restriction as a new type.

gears/src/x/ante.rs
```diff
@@ -142,6 +147,70 @@ impl<AK: AuthKeeper<SK>, BK: BankKeeper<SK>, SK: StoreKey, GC: SignGasConsumer>
...
+
+        if let Some(fee_coins) = fee {
+            let mut required_fees = Vec::with_capacity(min_gas_prices.len());
+
+            for gp in min_gas_prices {
+                required_fees.push(Coin {
+                    denom: gp.denom,
+                    amount: gp
+                        .amount
+                        .checked_mul(Decimal256::new(gas.into()))?
```
joneskm, line 180: Are you sure `Decimal256::new()` behaves correctly here e.g. is `Decimal256::new(1)` "equal" to 0.000000000000000001?

gears/src/x/ante.rs
```diff
@@ -142,6 +147,70 @@ impl<AK: AuthKeeper<SK>, BK: BankKeeper<SK>, SK: StoreKey, GC: SignGasConsumer>
...
+                        .to_uint_ceil(),
+                });
+            }
+
+            if !is_any_gte(fee_coins.inner(), &required_fees) {
```
joneskm, line 185: Does this mean that if one fee is greater than the required fee then the tx passes?

NYBACHOK: Right. This code from cosmos 

```go
if !feeCoins.IsAnyGTE(requiredFees) {
  return ctx, sdkerrors.Wrapf(sdkerrors.ErrInsufficientFee, "insufficient fees; got: %s required: %s", feeCoins, requiredFees)
}
```

gears/src/types/base/coin.rs
NYBACHOK: You right. I know about this restriction, and `.collect::<HashMap<_, _>>();` "resolves" this, but the question is what behavior should be.  Return error, panic or ignore as it does now

gears/src/x/ante.rs
NYBACHOK: Yes

```rust
#[test]
fn test() {
    let dec = Decimal256::new(Uint256::one());

    assert_eq!(dec, Decimal256::from_str("0.000000000000000001").unwrap())
}
```

gears/src/x/ante.rs
NYBACHOK: Isn't this what I implemented? If required fees(min_gas_prices?) is zero or empty we exit from method.

gears/src/baseapp/options.rs
NYBACHOK: Added a newtype for it

gears/src/types/base/coin.rs
NYBACHOK: Edit: see comment below

gears/src/x/ante.rs
joneskm: You're right I missed the check above where you exit.

gears/src/x/ante.rs
joneskm: Ah I just realized this is the expected behaviour.

gears/src/x/ante.rs
joneskm: OK but is this how it is in Cosmos? I expected 1 unit of gas to "equal" 1, not 0.00000000000000001.

NYBACHOK: Judging from [these](https://github.com/cosmos/cosmos-sdk/blob/d3f09c222243bb3da3464969f0366330dcb977a8/types/decimal_test.go#L97) tests 1 dec is 1.0000...

```go
func (s *decimalTestSuite) TestDecString() {
	tests := []struct {
		d    sdk.Dec
		want string
	}{
		{sdk.NewDec(0), "0.000000000000000000"},
		{sdk.NewDec(1), "1.000000000000000000"},
		{sdk.NewDec(10), "10.000000000000000000"},
		{sdk.NewDec(12340), "12340.000000000000000000"},
		{sdk.NewDecWithPrec(12340, 4), "1.234000000000000000"},
		{sdk.NewDecWithPrec(12340, 5), "0.123400000000000000"},
		{sdk.NewDecWithPrec(12340, 8), "0.000123400000000000"},
		{sdk.NewDecWithPrec(1009009009009009009, 17), "10.090090090090090090"},
	}
	for tcIndex, tc := range tests {
		s.Require().Equal(tc.want, tc.d.String(), "bad String(), index: %v", tcIndex)
	}
}
```

joneskm: Fixed in subsequent commit

gears/src/types/base/min_gas.rs
```diff
@@ -0,0 +1,111 @@
...
+use serde::{Deserialize, Serialize};
+
+use super::{
+    decimal_coin::DecimalCoin,
+    errors::{CoinsParseError, SendCoinsError},
+};
+
+// Represents a list of coins with the following properties:
+// - Contains at least one coin
+// - All coin amounts are positive
```
joneskm, line 12: Min gas can be zero

NYBACHOK: Resolved

PR #92, opened 2024-05-21
Gas metering for KVStore

NYBACHOK:


gears/src/params/mod.rs
```diff
@@ -1,46 +1,45 @@
...
+    CTX: InfallibleContextMut<DB, SK>,
     PSK: ParamsSubspaceKey,
 >(
     ctx: &'a mut CTX,
     params_subspace_key: &PSK,
 ) -> ParamsSpaceMut<'a, PrefixDB<DB>> {
     ParamsSpaceMut {
-        inner: ctx
-            .kv_store_mut(SK::params())
+        inner: InfallibleContextMut::infallible_store_mut(ctx, SK::params())
```
dmvict, line 42: It looks strange. Can you explain it?

gears/src/context/mod.rs
```diff
@@ -10,20 +12,31 @@ pub mod tx;
 
 pub trait QueryableContext<DB, SK> {
     /// Fetches an immutable ref to a KVStore from the MultiStore.
-    fn kv_store(&self, store_key: &SK) -> KVStore<'_, PrefixDB<DB>>;
+    fn kv_store(&self, store_key: &SK) -> Store<'_, PrefixDB<DB>>;
```
dmvict, line 15: As I understand you've used the method `kv_store` and `infallible_store` in params keepers. 

Where we can use it except params? Can we rewrite this part?

gears/src/params/mod.rs
NYBACHOK: I went with 2 different params stores. One is `Infallible` and use Init, Block and Query contexts, but there alternative to it which uses fallible version of context API

gears/src/context/mod.rs
NYBACHOK: Not quite understand this question. `KvStore` is mainly used for params.

PR #102, opened 2024-05-26
Add health and tx service

joneskm:


x/auth/src/client/grpc.rs
```diff
@@ -41,6 +42,7 @@ where
         &self,
         request: Request<QueryAccountRequest>,
     ) -> Result<Response<QueryAccountResponse>, Status> {
+        info!("Received a gRPC request auth::account");
```
NYBACHOK, line 45: I think that there should be any way to omit this. Maybe write some middleware for server

NYBACHOK (approved):

PR #104, opened 2024-05-28
Various tidying up

joneskm:


gears/src/baseapp/abci.rs
```diff
@@ -42,50 +40,28 @@ use tendermint::{
...
-            error!("Invalid chain id provided by Tendermint.\nTerminating process\n");
-            std::process::exit(1)
-        });
-
         let mut ctx = InitContext::new(
             &mut multi_store,
             self.block_height(),
             request.time.clone(),
-            chain_id.clone(),
+            request.chain_id.clone(),
```
NYBACHOK, line 57: Is it possible to remove this cloning?

tendermint/src/application.rs
```diff
@@ -37,7 +38,7 @@ use crate::types::{
 /// application.
 ///
 /// [`Server`]: crate::Server
-pub trait ABCIApplication: Send + Clone + 'static {
+pub trait ABCIApplication<G>: Send + Clone + 'static {
```
dmvict, line 41: It's a question. 
Why do we need this generic in declaration?

It seems like the `gaia-rs` won't compile after the change.

dmvict (approved):

NYBACHOK (approved):

gears/src/baseapp/abci.rs
joneskm: Yes! Pushed changes.

PR #105, opened 2024-05-31
chore: implement delegation and redelegation commands 

dmvict:


x/staking/src/keys.rs
```diff
@@ -0,0 +1,19 @@
+use gears::types::address::BaseAddress;
+
+pub fn length_prefixed_addr_key<const PREFIX: u8>(addr: &BaseAddress<PREFIX>) -> Vec<u8> {
```
NYBACHOK, line 3: What issue with usage of `AccAdress` or `ValAdress`?

NYBACHOK (approved):
See no issue with PR, but beg you to merge it after my PR

x/staking/src/types/query.rs
```diff
@@ -41,11 +54,122 @@ impl From<QueryValidatorRequest> for QueryValidatorRequestRaw {
...
+    }
+}
+
+impl Protobuf<QueryDelegationRequestRaw> for QueryDelegationRequest {}
+
+/// QueryRedelegationRequest is request type for the Query/Redelegation RPC method.
+#[derive(Clone, Debug, PartialEq)]
+pub struct QueryRedelegationRequest {
+    /// delegator_addr defines the delegator address to query for.
+    pub delegator_address: Option<AccAddress>,
```
joneskm, line 105: Should these be required rather than optional?

x/staking/src/keys.rs
```diff
@@ -0,0 +1,19 @@
...
+    let addr_str = addr.to_string();
+    let bytes = addr_str.as_bytes();
+    let mut bytes_prefix = bytes.len().to_ne_bytes().to_vec();
```
joneskm, line 6: `to_ne_bytes` isn't deterministic across nodes - a different result may be returned on different machines. Is this what the Cosmos SDK does?

x/staking/src/types/domain.rs
```diff
@@ -256,6 +265,56 @@ impl Validator {
...
 
+    pub fn shares_from_tokens_truncated(&self, amount: Uint256) -> anyhow::Result<Decimal256> {
+        if self.tokens.is_zero() {
+            return Err(AppError::Custom("insufficient shares".into()).into());
+        }
+
+        // TODO: check
+        let mul = self.delegator_shares.checked_mul(Decimal256::new(amount))?;
+        // TODO: check constant 18 in decimals
+        let precision_reuse = Decimal256::new(Uint256::from_u128(10u128)).checked_pow(18)?;
```
joneskm, line 276: `Decimal256::new(x)` has slightly unexpected behaviour e.g. `Decimal256::new(1)` "=" 0.000000000000000001. Is this the behaviour you want here?

x/staking/src/types/domain.rs
```diff
@@ -328,3 +387,82 @@ impl Validator {
...
+    fn try_from(value: ValidatorRaw) -> Result<Self, Self::Error> {
+        let status = value.status();
+        Ok(Self {
+            operator_address: ValAddress::from_bech32(&value.operator_address)
+                .map_err(|e| Error::DecodeAddress(e.to_string()))?,
+            delegator_shares: Decimal256::from_str(&value.delegator_shares)
+                .map_err(|e| Error::DecodeGeneral(e.to_string()))?,
+            description: value
+                .description
+                .expect("Value should exists. It's the proto3 rule to have Option<T> instead of T"),
```
joneskm, line 402: Is this guaranteed not to panic? Perhaps return an error?

x/staking/src/types/tx_proto.rs
```diff
@@ -72,6 +72,31 @@ pub struct Commission {
...
 
+impl TryFrom<CommissionRaw> for Commission {
+    type Error = Error;
+
+    fn try_from(value: CommissionRaw) -> Result<Self, Self::Error> {
+        Ok(Self {
+            commission_rates: value
+                .commission_rates
+                .ok_or(Error::MissingField(
+                    "Value should exists. It's the proto3 rule to have Option<T> instead of T"
```
joneskm, line 83: I'm not sure a user would understand this message. Something like "missing field X" might be better?

x/staking/src/types/domain.rs
NYBACHOK: If you want 1 use `Decimal256::from_atomics(1, 0)`

x/staking/src/types/domain.rs
NYBACHOK: Small comment. proto3 doesn't have rule for all fields to be optional if you meant this. It makes all fields optional by default

x/staking/src/keys.rs
dmvict: This handler is universal for any type of account addresses and reused from the routine [`length_prefixed_addr_pair_key`](https://github.com/rumos-io/gears/pull/105/files#diff-8251c004799364f46846eb7e3d366412d1be013eb3d61563fe57a52956f553a2R11). I used it twice in the routine: for an AccAddress and for a ValAddress. And if I'm correct, I used it once in the other method. 

The better question is: 
Why do you use generics in routine `length_prefixed_addr_pair_key`? 

I'm answering that the sdk has many methods for concrete types and it's a mess of code. This methods create keys for stores. 

I can convert the method `length_prefixed_addr_pair_key` to another signature and I'll rename it. It makes sense for now. I don't want to change the `length_prefixed_addr_key` because it is very simple and straightforward routine. 

@NYBACHOK What do you think about it?

NYBACHOK: I was only curious about usage of this method. If you say that you reuse this method for 2 different addresses, then see no issue with keeping this one method with generic.

x/staking/src/types/query.rs
dmvict: The struct can have type without `Option`. I used this signature that is close to sdk signature because it allows to have one node side handler for a few queries. [Here](https://github.com/rumos-io/gears/pull/105/files#diff-56b566899aff451a0f8dcccb07a4c0906dc4952ff18c9fdf17d5839adc234fdaR70) you can find match statement for the queries. 

If I'll change the type, then I need to have several additional structures for queries and additional handlers. It's just another solution. 

Should I change the code?

x/staking/src/keys.rs
dmvict: It doesn't. I'll change to `to_le_bytes`.

x/staking/src/types/domain.rs
dmvict: I expected another behavior. I'll check the code and fix the places with such constructors.

NYBACHOK: It was unexpected to me, too.

x/staking/src/types/domain.rs
dmvict: The protobuf conversion says that messages aka serialized structs are serialized to `Option<T>`. Please, look at [this section](https://docs.rs/prost/latest/prost/#field-modifiers). 
Since I wrote serialization from original type to raw version I expect that the protobuf cannot miss field of original type. It's like `unwrap_or_corrupt`. 

If you disagree with this thoughts I'll update the code.

x/staking/src/types/tx_proto.rs
dmvict: Thank you, I'll update the messages.

x/staking/src/keys.rs
joneskm: I think you've found a flaw in the way we handle addresses in Gears. It was not the intention that users could access the `BaseAddress`. In future we'll change the gears implementation so that you can't. 
Do you need the flexibility here? Where it's used the first address is a `ValAddress` and the second is an `AccAddress`. If you need the flexibility then perhaps pass in the raw bytes or define your own enum. Perhaps in future we''ll have an `Address` trait in gears which you could use.

x/staking/src/types/query.rs
joneskm: If you can still handle a query (i.e. without returning an error) without a field, then it should be optional.
If I understand you correctly, you're re-using the struct for several queries. For some queries some fields are optional while for other queries they're required. Is that right? In which case go with whatever you prefer.

x/staking/src/types/domain.rs
joneskm: Only the code calling this method knows whether to `expect` these fields to be present e.g. if the calling code is reading serialized bytes from state then that code can `expect` the fields to be present. If however the calling code is deserializing a Transaction then that code would need handle the error case (you can't guarantee that the submitted transaction was created using your code).
TLDR: I don't think you can safely unwrap here since you don't have enough context.

x/staking/src/types/query.rs
dmvict: >  For some queries some fields are optional while for other queries they're required. Is that right? In which case go with whatever you prefer.

Yes, it's right. The struct is used for several queries and the client code defines required/optional fields. I prefer to have one struct for redelegation queries because the other structs are the subsets of the current one.

x/staking/src/types/domain.rs
```diff
@@ -235,7 +244,8 @@ impl Validator {
         // calculate the shares to issue
         let issues_shares = if self.delegator_shares.is_zero() {
             // the first delegation to a validator sets the exchange rate to one
-            Decimal256::new(amount)
+            Decimal256::from_atomics(amount, 0)
+                .expect("Conversion won't fail. The number of decimal places is 0")
```
joneskm, line 248: I don't think it's true that this conversion can't fail see https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.Decimal256.html "The greatest possible value that can be represented is 115792089237316195423570985008687907853269984665640564039457.584007913129639935 (which is (2^256 - 1) / 10^18)"

PR #106, opened 2024-06-03
Refactor execution modes

NYBACHOK:


gears/src/baseapp/mod.rs
```diff
@@ -195,7 +193,10 @@ impl<PSK: ParamsSubspaceKey, H: ABCIHandler, AI: ApplicationInfo> BaseApp<PSK, H
         ctx.block_gas_meter
             .consume_gas(gas_used, BLOCK_GAS_DESCRIPTOR)?;
 
-        MD::commit(ctx, &mut multi_store);
+        std::mem::drop(ctx);
```
dmvict, line 196: Just a question. 
Why do you need to drop it here?

dmvict (approved):

gears/src/baseapp/mod.rs
NYBACHOK: To release mutable reference of `MultiBank<DB, AH::StoreKey, TransactionStore>` from ctx

NYBACHOK:
@joneskm resolved conflicts

NYBACHOK:
@joneskm resolved issue again

PR #107, opened 2024-06-03
Move rocksb to gaia

NYBACHOK:


database/src/memory.rs
```diff
@@ -1,10 +1,14 @@
...
+    ops::Bound,
+    sync::{Arc, RwLock},
+};
 
 use crate::Database;
 
 #[derive(Debug, Clone)]
 pub struct MemDB {
-    store: RefCell<BTreeMap<Vec<u8>, Vec<u8>>>, // we use a refcell because the set method on the DB trait doesn't take a mutable ref
+    store: Arc<RwLock<BTreeMap<Vec<u8>, Vec<u8>>>>, // we use a refcell because the set method on the DB trait doesn't take a mutable ref
```
dmvict, line 11: Update/remove comment, please.

gears/src/commands/node/run.rs
```diff
@@ -32,7 +32,7 @@ pub enum RunError {
     #[error("{0}")]
     HomeDirectory(String),
     #[error("{0}")]
-    Database(#[from] database::error::Error),
+    Database(String),
```
dmvict, line 35: Why did you replaced the type to `String`?

database/src/memory.rs
NYBACHOK: Done

gears/src/commands/node/run.rs
NYBACHOK: Error type under `rockdb` feature

database/src/lib.rs
```diff
@@ -23,3 +23,9 @@ pub trait Database {
         prefix: Vec<u8>,
     ) -> Box<dyn Iterator<Item = (Box<[u8]>, Box<[u8]>)> + 'a>;
 }
+
+pub trait DatabaseBuilder<DB> {
```
joneskm, line 27: Would it make more sense for the DB to be an associated type?

NYBACHOK: I thought about it, but think it's easier to use simple generic + generic allows different implementations.
I mean that you could use single structure for different databases.

joneskm: Fair enough.

PR #108, opened 2024-06-03
Small cleanup

joneskm:
The unreachable code was reachable!

### DIFF
--- a/gears/src/baseapp/mod.rs
+++ b/gears/src/baseapp/mod.rs
@@ -69,7 +69,8 @@ impl<PSK: ParamsSubspaceKey, H: ABCIHandler, AI: ApplicationInfo> BaseApp<PSK, H
             params_subspace_key,
         };
 
-        let ctx = SimpleContext::new(&mut multi_store, 0);
+        let height = multi_store.head_version() as u64;
+        let ctx = SimpleContext::new(&mut multi_store, height);
 
         let max_gas = baseapp_params_keeper
             .block_params(&ctx)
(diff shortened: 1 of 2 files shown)


NYBACHOK:
1. how this even possible or better say where?
2. Why not hardcoded zero for height?
3. I think that this may be an issue and height should be ignored.

joneskm:
1. e.g. call `let height = ctx.height();` [here](https://github.com/rumos-io/gears/blob/59a9c52844efbc021eec0845a3b1c4cbcf6cdc60/gears/src/baseapp/mod.rs#L74)
2. We can use the actual height.
3. What do you think the issue might be?

NYBACHOK:
1. Check your URL. There is no such code
2. Ok
3. Hard to explain. It more about feeling that what piece of code need height? I don't remember where in params we use height. And the only reason for this context to exists is to read params from store when we create app. 
I don't think it's weird to use actual height, but found it strange

joneskm:
1. I'm saying you could put that code there, then the unreachable code is reached.
2. I see what you mean, but the reality is it does expose a height method so it's best that it returns a "sensible" height.

NYBACHOK:
If it worries you then see no issue. Last thing, I would ask why you use version of multi store not app block height

joneskm:
In `new`? The app block height hasn't been set at that point. In fact it gets set to the same height a few lines later.

NYBACHOK:
Not only. We [use](https://github.com/NYBACHOK/gears/blob/8639e90b65e0e531dfa950245d5ba3b2399a7db8/gears/src/baseapp/mod.rs#L163-L167) `SimpleContext` to read consensus params

joneskm:
There we do use the app block height.

NYBACHOK:
I get it now. There no other "issue"

PR #110, opened 2024-06-04
Init genesis in gov xmod

NYBACHOK:
Part of https://github.com/rumos-io/gears/issues/117


x/gov/src/keeper.rs
```diff
@@ -0,0 +1,136 @@
...
+           Okay. I think that in our implementation there is no need to create account if it.
+
+           So I should omit this lines...
+           if balance.is_empty() || balance.iter().any(|this| this.amount.is_zero()) {
+               https://github.com/cosmos/cosmos-sdk/blob/d3f09c222243bb3da3464969f0366330dcb977a8/x/gov/genesis.go#L47
+           }
+        */
+
+        if !(balance == total_deposits) {
+            panic!(
```
joneskm, line 130: So we're expecting that in the bank module init genesis the gov_mod account will be initialized to the right amount?

x/gov/src/types/deposit.rs
```diff
@@ -0,0 +1,23 @@
...
+    pub depositor: AccAddress,
+    pub amount: Vec<Coin>,
+}
+
+impl Deposit {
+    pub(crate) fn key(&self) -> Vec<u8> {
+        [
+            KEY_DEPOSIT_PREFIX.as_slice(),
+            &self.proposal_id.to_be_bytes(),
+            &[self.depositor.len() as u8], // We save 'cause `AccAddress` len shoudn't be bigger than 255
```
joneskm, line 18: len() returns a u8, so we don't need the cast

x/gov/src/types/proposal.rs
```diff
@@ -0,0 +1,62 @@
...
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct Proposal {
+    pub proposal_id: u64,
+    pub content: Vec<u8>, // TODO
+    pub status: ProposalStatus,
+    pub final_tally_result: (), // TODO: https://github.com/cosmos/cosmos-sdk/blob/d3f09c222243bb3da3464969f0366330dcb977a8/x/gov/types/gov.pb.go#L289
+    pub submit_time: (),
+    pub deposit_end_time: DateTime<Utc>,
+    pub total_deposit: Vec<Coin>,
```
joneskm, line 21: Are there any additional constraints on the list of coins? Do we need a new type?

x/gov/src/types/vote.rs
```diff
@@ -0,0 +1,23 @@
...
+    pub voter: AccAddress,
+    pub options: (), // TODO:
+}
+
+impl Vote {
+    pub fn key(&self) -> Vec<u8> {
+        [
+            KEY_VOTES_PREFIX.as_slice(),
+            &self.proposal_id.to_be_bytes(),
+            &[self.voter.len() as u8], // We save 'cause `AccAddress` len shoudn't be bigger than 255
```
joneskm, line 18: Same as before. len() returns a u8.

x/gov/src/types/deposit.rs
NYBACHOK: Unexpected for me...

x/gov/src/types/proposal.rs
NYBACHOK: For now, it's okay, but later maybe I add a new type or use `SendCoins`( it needs to be renamed)

x/gov/src/keeper.rs
NYBACHOK: Right. But I need to make sure that it created before usage(I suspected that It doesn't) and additional thing to note is denom.

x/gov/src/types/deposit.rs
NYBACHOK: Removed

x/gov/src/types/vote.rs
NYBACHOK: Removed

x/gov/src/types/proposal.rs
joneskm: Fair enough. I agree with the renaming, `Vec<Coin>` seems to be appearing more and more.

PR #111, opened 2024-06-04
chore: implement method `begin_block`

dmvict:


x/staking/src/keeper/validators_and_total_power.rs
```diff
@@ -96,6 +98,60 @@ impl<
...
 
+    /// get the last validator set
+    pub fn last_validators_by_addr<DB: Database, CTX: QueryableContext<DB, SK>>(
+        &self,
+        ctx: &CTX,
+    ) -> HashMap<String, Vec<u8>> {
+        let mut last = HashMap::new();
+        let store = ctx.kv_store(&self.store_key);
+        let store = store.prefix_store(LAST_VALIDATOR_POWER_KEY);
+        for (k, v) in store.range(..).to_infallible_iter() {
```
joneskm, line 109: @NYBACHOK this is why I think we should get rid of `to_infallible_iter`.

x/staking/src/keeper/historical_info.rs
```diff
@@ -0,0 +1,96 @@
...
+        // In the rare scenario when the historical entries gets reduced to a lower value k'
+        // from the original value k. k - k' entries must be deleted from the store.
+        // Since the entries to be deleted are always in a continuous range, we can iterate
+        // over the historical entries starting from the most recent version to be pruned
+        // and then return at the first empty entry.
+
+        if (ctx.height() as i64 - entry_num as i64) >= 0 {
+            for i in (ctx.height() - entry_num as u64)..=0 {
+                // SAFETY: the BlockContext is infallible context in context of StoreErrors
+                if let Some(_info) = self.historical_info(ctx, i).unwrap() {
```
joneskm, line 29: @NYBACHOK I think we should have a `GasError` on the store method and have add an `unwrap_gas` method to `Result`

x/staking/src/keeper/validators_and_total_power.rs
NYBACHOK: You mean @dmvict overuses it? I don't think absolutely good idea...

x/staking/src/keeper/historical_info.rs
```diff
@@ -0,0 +1,96 @@
...
+
+        // Prune store to ensure we only have parameter-defined historical entries.
+        // In most cases, this will involve removing a single historical entry.
+        // In the rare scenario when the historical entries gets reduced to a lower value k'
+        // from the original value k. k - k' entries must be deleted from the store.
+        // Since the entries to be deleted are always in a continuous range, we can iterate
+        // over the historical entries starting from the most recent version to be pruned
+        // and then return at the first empty entry.
+
+        if (ctx.height() as i64 - entry_num as i64) >= 0 {
```
joneskm, line 26: This casting will cause wrap-around issues

x/staking/src/keeper/validators_and_total_power.rs
NYBACHOK: Maybe something like `#[deprecated]` will do the trick to better inform user

x/staking/src/keeper/historical_info.rs
NYBACHOK: What it gives for us? Unified unwrap in case of infallible ctx?

x/staking/src/keeper/historical_info.rs
```diff
@@ -0,0 +1,93 @@
...
+
+        // Prune store to ensure we only have parameter-defined historical entries.
+        // In most cases, this will involve removing a single historical entry.
+        // In the rare scenario when the historical entries gets reduced to a lower value k'
+        // from the original value k. k - k' entries must be deleted from the store.
+        // Since the entries to be deleted are always in a continuous range, we can iterate
+        // over the historical entries starting from the most recent version to be pruned
+        // and then return at the first empty entry.
+
+        if !ctx.height().overflowing_sub(entry_num as u64).1 {
```
joneskm, line 25: Can this line be simplified to `if ctx.height() >= entry_num as u64 {'?

PR #113, opened 2024-06-04
Use u32 for block heights

joneskm:
Since Tendermint uses i64 to represent heights the only sensible option is to use a u32 in gears.

### DIFF
--- a/gears/src/types/tx/body.rs
+++ b/gears/src/types/tx/body.rs
@@ -29,7 +29,7 @@ pub struct TxBody<M> {
     /// timeout is the block height after which this transaction will not
     /// be processed by the chain
     #[serde_as(as = "DisplayFromStr")]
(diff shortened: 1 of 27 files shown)


gears/src/types/tx/body.rs
```diff
@@ -73,7 +79,7 @@ impl<M: TxMessage> From<TxBody<M>> for inner::TxBody {
                 .map(|this| this.into())
                 .collect(),
             memo: tx_body.memo,
-            timeout_height: tx_body.timeout_height,
+            timeout_height: tx_body.timeout_height as u64, //TODO: consider using a copy of the raw TxBody struct which has a u32 timeout_height
```
NYBACHOK, line 82: You may use it, but only if you need to change other fields too. It absolutely fine

PR #116, opened 2024-06-12
chore: integrate `staking` into gaia 

dmvict:


x/auth/src/lib.rs
```diff
@@ -11,3 +11,16 @@ pub use genesis::*;
 pub use keeper::*;
 pub use message::*;
 pub use params::*;
+
+//
+
+pub fn new_module_addr(module_name: &str) -> gears::types::address::AccAddress {
```
NYBACHOK, line 17: You may use [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html)

gaia-rs/src/modules.rs
```diff
@@ -21,12 +25,16 @@ impl Module for GaiaModules {
...
             }
+            GaiaModules::BondedPool => auth::new_module_addr(&self.get_name()),
+            GaiaModules::NotBondedPool => auth::new_module_addr(&self.get_name()),
         }
     }
 
     fn get_permissions(&self) -> Vec<String> {
         match self {
             GaiaModules::FeeCollector => vec![],
+            GaiaModules::BondedPool => vec!["staking".into()],
```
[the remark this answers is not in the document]
dmvict: It's compatibility thing. You're right, it's better to use enum.

x/auth/src/lib.rs
dmvict: I've updated realization but I cannot use OnceLock because I need to have a mutable instance for creating a hash. The new solution is a little better because I used a special method.

PR #119, opened 2024-06-18
Implement keeper for gov xmod

NYBACHOK:


Cargo.toml
```diff
@@ -13,7 +13,7 @@ members = [
   "gears-derive",
 
   # GAIA
-  "gaia-rs",
+  # "gaia-rs",
```
joneskm: Uncomment!

dmvict, line 16: Why do you do this?

x/bank/Cargo.toml
```diff
@@ -9,7 +9,7 @@ gears = { path = "../../gears", features = ["cli", "export", "xmods"] }
 
 #newtypes
 bytes = { workspace = true }
-ibc-proto = { workspace = true } # TODO: should we import through gears via core-types?
+ibc-proto = { workspace = true }
```
dmvict, line 12: Why did you remove this?

x/gov/src/keeper.rs
```diff
@@ -133,4 +144,271 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, M: Module, BM: Module, BK: BankKeeper
...
+        None => Ok(None),
+    }
+}
+
+fn deposit_set<DB: Database>(
+    mut store: GasKVStoreMut<'_, DB>,
+    deposit: &MsgDeposit,
+) -> Result<(), GasStoreErrors> {
+    store.set(
+        MsgDeposit::key(deposit.proposal_id, &deposit.depositor),
```
dmvict, line 379: Can you update signature? It would look like `deposit.generete_key()`.

x/gov/src/msg/proposal.rs
```diff
@@ -0,0 +1,119 @@
...
+        Ok(MsgSubmitProposal::decode::<Bytes>(value.value.into())
+            .map_err(|e| CoreError::DecodeProtobuf(e.to_string()))?)
+    }
+}
+
+impl From<MsgSubmitProposal> for Any {
+    fn from(msg: MsgSubmitProposal) -> Self {
+        Any {
+            type_url: MsgSubmitProposal::TYPE_URL.to_string(),
+            value: msg.encode_vec().expect(IBC_ENCODE_UNWRAP),
```
dmvict, line 116: If I'm correct, the newer protobuf converts  bytes without result.

x/gov/src/msg/vote.rs
```diff
@@ -0,0 +1,132 @@
...
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct MsgVote {
+    pub proposal_id: u64,
+    pub voter: AccAddress,
+    pub option: VoteOption,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum VoteOption {
+    Empty = 0,
```
dmvict, line 26: Please, use common format for all options.

x/gov/src/keeper.rs
```diff
@@ -133,4 +144,271 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, M: Module, BM: Module, BK: BankKeeper
...
+        ctx: &mut TxContext<'_, DB, SK>,
+        MsgSubmitProposal {
+            content,
+            initial_deposit,
+            proposer: _proposer,
+        }: MsgSubmitProposal,
+    ) -> anyhow::Result<u64> {
+        /*
+           in go they perform check is it possible to handle
+           proposal somehow, but not sure we need it and instead
```
joneskm, line 265: Perhaps add a link to the cosmos code?

x/bank/Cargo.toml
NYBACHOK: We had this discussion, or at least I think that we had. We need to export types in gears only if they being used more than once. So this comment is outdated.

NYBACHOK: Maybe better example why we should omit re-export of structures is `ibc-rs` xmod which rely on different version of `ibc-proto` then `gears`

x/gov/src/msg/proposal.rs
NYBACHOK: I use older from gears here.

x/gov/src/msg/vote.rs
NYBACHOK: Not sure what you mean?

Cargo.toml
NYBACHOK: Done

x/gov/src/keeper.rs
NYBACHOK: Done

x/gov/src/msg/vote.rs
dmvict: It is better to skip byte representation or add it for each option. 

I know that below contains code with this description. Some formal rules improves reading and it can.

NYBACHOK: Added to every option

PR #121, opened 2024-06-18
chore: add integration tests for staking module inside `gaia-rs`

dmvict:


x/staking/src/traits.rs
```diff
@@ -23,7 +23,7 @@ pub trait BankKeeper<SK: StoreKey, M: Module>: Clone + Send + Sync + 'static {
 
     fn get_all_balances<DB: Database, CTX: QueryableContext<DB, SK>>(
         &self,
-        ctx: &mut CTX, //TODO: why is this mutable?
```
NYBACHOK, line 26: It was mutable 'cause we used concrete context, which always was mutable

gaia-rs/tests/staking.rs
```diff
@@ -0,0 +1,490 @@
...
+    BondStatus, CommissionRatesRaw, CommissionRaw, DelegationResponse, Description,
+    RedelegationResponse, Validator,
+};
+use std::{path::PathBuf, str::FromStr};
+use utilities::ACC_ADDRESS;
+
+#[path = "./utilities.rs"]
+mod utilities;
+
+fn run_tx_local(
```
NYBACHOK, line 40: Could we reuse this method in others tests?

gaia-rs/tests/staking.rs
```diff
@@ -0,0 +1,490 @@
...
+            node: DEFAULT_TENDERMINT_RPC_ADDRESS.parse()?,
+            chain_id: ChainId::from_str("test-chain")?,
+            fee: None,
+            inner: WrappedGaiaTxCommands(command),
+        },
+        &GaiaCoreClient,
+    )
+}
+
+fn run_query_local(command: GaiaQueryCommands) -> anyhow::Result<GaiaQueryResponse> {
```
NYBACHOK, line 61: Same here

NYBACHOK (approved):

gaia-rs/tests/staking.rs
dmvict: Do you mean `utilities.rs`? Of course, we can. I suggest  it's  a smaller abstraction that allows to reduce a lot of code. 

I've used it many times in the file.

NYBACHOK: May you use it in this PR? Or make another one

PR #123, opened 2024-06-21
chore: improve staking, add methods, resolve issues

dmvict:


x/staking/src/keeper/redelegation.rs
```diff
@@ -348,7 +343,7 @@ impl<
         };
 
         let storage = InfallibleContextMut::infallible_store_mut(ctx, &self.store_key);
-        let mut store = storage.prefix_store_mut(UNBONDING_QUEUE_KEY);
+        let mut store = storage.prefix_store_mut(REDELEGATION_QUEUE_KEY);
```
NYBACHOK, line 346: Did you only rename key or changed it value? If you did, then check state hash of main and your branch

x/staking/src/keeper/store_iter.rs
```diff
@@ -0,0 +1,63 @@
...
+    store::database::{ext::UnwrapCorrupt, prefix::PrefixDB},
+    types::{
+        address::ValAddress,
+        store::{gas::errors::GasStoreErrors, kv::Store, range::StoreRange},
+    },
+};
+use std::{borrow::Cow, ops::Bound};
+
+#[derive(Debug)]
+struct StoreIterator<'a, DB>(StoreRange<'a, PrefixDB<DB>>);
```
NYBACHOK, line 12: Maybe we need this structure in gears?

x/staking/src/keeper/store_iter.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+impl<'a, DB: Database> Iterator for UnbondingValidatorsIterator<'a, DB> {
+    type Item = Result<(Vec<u8>, Vec<ValAddress>), GasStoreErrors>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0
+            .next()
+            .map(|r| r.map(|(k, v)| (k.to_vec(), serde_json::from_slice(&v).unwrap_or_corrupt())))
```
NYBACHOK, line 61: Not sure that `unwrap_or_corrupt` is correct in this case

x/staking/src/types/domain.rs
```diff
@@ -275,7 +279,7 @@ impl Validator {
             // TODO: infallible in sdk
             Decimal256::from_atomics(amount, 0).unwrap()
         } else {
-            // TODO: check the code, maybe remove unwrap
+            // panics in original implementation
```
NYBACHOK, line 282: Don't think that we need to follow cosmos logic on panic in gears. Remember gas case

x/staking/src/types/domain.rs
```diff
@@ -323,11 +324,9 @@ impl Validator {
         } else {
             // leave excess tokens in the validator
             // however fully use all the delegator shares
-            // TODO: infallible + floor
             let tokens = self.tokens_from_shares(del_shares).unwrap().to_uint_floor();
-            // TODO: check of negative result
+            // the library panics on substruct with overflow and this behavior is identical to sdk
```
NYBACHOK, line 328: Same comment about panic

tendermint/src/types/proto/crypto.rs
```diff
@@ -56,6 +58,14 @@ impl TryFrom<inner::PublicKey> for PublicKey {
     }
 }
 
+impl FromStr for PublicKey {
```
NYBACHOK, line 61: Not sure that this is a correct way of how `FromStr` should be implemented.

x/staking/src/keeper/redelegation.rs
dmvict: It was wrong key there.

I'll check.

x/staking/src/keeper/store_iter.rs
dmvict: We can have such structs. One for infallible stores, the other one for fallible.

x/staking/src/keeper/store_iter.rs
dmvict: It deserializes data from a store for a concrete domain type. I guess that it is correct place.
I had the same code but without iterator.

x/staking/src/keeper/store_iter.rs
NYBACHOK: If you want

tendermint/src/types/proto/crypto.rs
dmvict: The cosmos cli uses json based string to pass the key. And it's implementation for described case. 

Maybe, we can skip a part with key type declaration and hardcode `Ed25519`. Then we can read string with only key data.

x/staking/src/keeper/store_iter.rs
NYBACHOK: `unwrap_or_corrupt` panic with message that value expected to be in DB, but it wasn't found

tendermint/src/types/proto/crypto.rs
NYBACHOK: Okay. I get it now. You implemented `FromStr` 'cause of `clap`?

x/staking/src/types/domain.rs
dmvict: @joneskm What do you think? 
Should we panic if we cannot create/update a coin/share?

I suggest that the panic is correct because it locks node and it cannot update the chain with wrong amount of coins.

tendermint/src/types/proto/crypto.rs
NYBACHOK: I'll remind you that in our implementation cli command and command expected to be different structures, but which implement `TryFrom<T, Error = anyhow::Error>` and where I expected from user to do such conversions

x/staking/src/types/domain.rs
NYBACHOK: I have similar case... but it used in `end_block` which can't return errors

x/staking/src/keeper/store_iter.rs
dmvict: I've used a special message for such conversions but we decided to use `unwrap_or_corrupt`.  I can add the constant and use it again.

tendermint/src/types/proto/crypto.rs
dmvict: > Okay. I get it now. You implemented `FromStr` 'cause of `clap`?

I depends on input format. It is json string for now.

x/staking/src/keeper/store_iter.rs
NYBACHOK: If so, then it better change const value in database crate

tendermint/src/types/proto/crypto.rs
dmvict: > I'll remind you that in our implementation cli command and command expected to be different structures, but which implement TryFrom<T, Error = anyhow::Error> and where I expected from user to do such conversions

I had such conversion. I was asked to improve the cli with automatic conversion of `PublicKey` type.

NYBACHOK: Why not use some type which wrapps string and tries to parse it into `PublicKey`?

x/staking/src/types/domain.rs
joneskm: I guess this should never happen unless there's a bug somewhere? In which case `expect` it. My view is, if there's something that doesn't add up, better to halt the chain.

NYBACHOK: I don't think that halting node due to an issue with (for example) a transaction for one account is too much. Better return error for this tx

joneskm: I depends... if the problem is caused by the chain being in an "impossible/unreachable" state, even if it was found to be in such a state during tx processing, then the chain should be halted.

x/staking/src/keeper/redelegation.rs
dmvict: It's okay.

x/staking/src/keeper/unbonding.rs
```diff
@@ -219,11 +225,9 @@ impl<
...
 
             if height < block_height && (time <= block_time) {
-                for addr in v {
-                    let val_addr = ValAddress::from_bech32(addr)
-                        .expect("Failed to parse stored ValAddress in validators queue. Validators queue map contains vector of string addresses that could be a valid ValAddress representation.");
+                for val_addr in v {
                     let mut validator = self
-                        .validator(ctx, &val_addr)?
+                        .validator(ctx, val_addr)?
                         .expect("validator in the unbonding queue was not found");
```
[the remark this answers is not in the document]
dmvict: The sdk code requires a validator in this point. All `*mature*` methods require existed validators,

x/staking/src/types/domain.rs
dmvict: I checked the method. It is called from fallible method. But the case is rather `impossible/unreachable` because a use can't have a negative balance.

x/staking/src/types/domain.rs
joneskm: In this case, I don't think we should panic. This action can legitimately fail e.g. if the amount to delegate cannot fit into a `Decimal256`.

PR #124, opened 2024-06-21
implement end_block for gov xmod

NYBACHOK:


x/gov/src/keeper.rs
```diff
@@ -314,9 +355,302 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, M: Module, BK: BankKeeper<SK, M>>
...
+                } else {
+                    deposit_refund(ctx, self).unwrap_gas();
+                }
+
+                if passes {
+                    // TODO: Handle proposal: https://github.com/cosmos/cosmos-sdk/blob/d3f09c222243bb3da3464969f0366330dcb977a8/x/gov/abci.go#L65-L83
+
+                    proposal.status = ProposalStatus::Passed;
+
+                    if true
```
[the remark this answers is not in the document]
NYBACHOK: Consider that this is pseudocode

x/gov/src/keeper.rs
```diff
@@ -314,9 +355,302 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, M: Module, BK: BankKeeper<SK, M>>
...
+                    // from cosmos: https://github.com/cosmos/cosmos-sdk/blob/d3f09c222243bb3da3464969f0366330dcb977a8/x/gov/keeper/tally.go#L51
+                    // There is no need to handle the special case that validator address equal to voter address.
+                    // Because voter's voting power will tally again even if there will deduct voter's voting power from validator.
+
+                    validator.delegator_deduction += delegation.shares(); // TODO: handle overflow?
+
+                    // delegation shares * bonded / total shares
+                    let voting_power = delegation
+                        .shares()
+                        .mul(Decimal256::new(validator.bounded_tokens))
```
dmvict, line 517: It seems like here can be mistake with multiplication or with decimal. Please, check it.

x/gov/src/keeper.rs
```diff
@@ -314,9 +355,302 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, M: Module, BK: BankKeeper<SK, M>>
...
+                        .div(validator.delegator_shares);
+
+                    for VoteOptionWeighted { option, weight } in &vote_options {
+                        let result_option = tally_results.get_mut(option);
+
+                        *result_option += voting_power * Decimal256::from(weight.clone());
```
dmvict, line 523: What is the type of `weight`?

x/gov/src/keeper.rs
```diff
@@ -314,9 +355,302 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, M: Module, BK: BankKeeper<SK, M>>
...
+
+        // If there is no staked coins, the proposal fails
+        if total_bonded_tokens.amount.is_zero() {
+            return Ok((false, false, tally_results.result()));
+        }
+
+        // If there is not enough quorum of votes, the proposal fails
+        let percent_voting = total_voting_power / Decimal256::new(total_bonded_tokens.amount);
+        if percent_voting < tally_params.quorum {
+            return Ok((false, true, tally_results.result()));
```
dmvict, line 572: It seems like the method is a getter? I suggest that here we have a conversion. If yes, then it is better to change `to_result` or `to_tally_result`. It is more readable variants.

x/gov/src/keeper.rs
NYBACHOK: `Uint256`

x/gov/src/keeper.rs
NYBACHOK: Renamed to `to_result`

x/gov/src/keeper.rs
NYBACHOK: You mean replace methods call with operators or...?

dmvict: If I'm correct, you've created a number that has value `n * 10 ^ (-18)`. But seems like you need to have a decimal representation of an integer. 
the other thing is the overflows. Can you get an overflow?

x/gov/src/keeper.rs
dmvict: Maybe I missed something. It seems like it is incorrect conversion. Please, check it.

NYBACHOK: I lied. It's wrapper around `Decimal256`. so no issue with it

x/gov/src/keeper.rs
NYBACHOK: > But seems like you need to have a decimal representation of an integer.
I forget about it completely. Even that I gave you a similar comment in one of your PR's. Remind me I need `from_atomics( bounded_tokens, 0)`?
I don't handle any overflow for now

gears/src/x/keepers/bank.rs
```diff
@@ -34,4 +42,18 @@ pub trait BankKeeper<SK: StoreKey, M: Module>: Clone + Send + Sync + 'static {
         ctx: &CTX,
         address: &AccAddress,
     ) -> Result<Vec<Coin>, GasStoreErrors>;
+
+    fn coins_burn<DB: Database, CTX: QueryableContext<DB, SK>>(
```
joneskm, line 46: The name suggests that this method burns coins, but it's called from a queryable context. Maybe it needs a better name?

gears/src/x/types/delegation.rs
```diff
@@ -0,0 +1,9 @@
+use cosmwasm_std::Decimal256;
+
+use crate::types::address::{AccAddress, ValAddress};
+
+pub trait StakingDelegation {
+    fn delegator(&self) -> &AccAddress;
+    fn validator(&self) -> &ValAddress;
+    fn shares(&self) -> &Decimal256;
```
joneskm, line 8: So the shares can be fractional? Is this what they use in Cosmos?

gears/src/x/types/validator.rs
```diff
@@ -0,0 +1,22 @@
+use cosmwasm_std::{Decimal256, Uint256};
+use prost::Enumeration;
+use serde::{Deserialize, Serialize};
+
+use crate::types::address::ValAddress;
+
+pub trait StakingValidator {
+    fn operator(&self) -> &ValAddress;
+    fn bonded_tokens(&self) -> &Uint256;
+    fn delegator_shares(&self) -> &Decimal256;
```
joneskm, line 10: Same as above.

gears/src/x/keepers/bank.rs
NYBACHOK: I didn't implement new bank methods yet. Probably I'll change it in next PR with actual implementation. For now this is placeholder

gears/src/x/types/delegation.rs
NYBACHOK: This is what we have

gears/src/x/types/validator.rs
NYBACHOK: Similar case. I just followed @dmvict structures, but instead of moving them into gears added trait with getters for fields that I need

PR #125, opened 2024-06-21
[POC]  Graceful shutdown

NYBACHOK:


tendermint/src/application.rs
```diff
@@ -110,109 +115,200 @@ pub trait ABCIApplication<G>: Send + Clone + 'static {
...
         request: tendermint_proto::abci::RequestCheckTx,
     ) -> tendermint_proto::abci::ResponseCheckTx {
-        T::check_tx(&self.handler, request.into()).into()
+        match self.token.if_cancelled_ctx() {
+            Some(ctx) => match ctx {
+                CancellationContext::GasOverflow => {
+                    ResponseCheckTx::error_with_gas_overflow().into()
+                }
+                _ => {
+                    self.token.panic_if_cancelled();
```
dmvict, line 201: What are the other cases here? It's a little unclear here.

NYBACHOK: Whatever you like. This code only to show possible usage for gas and maybe something else

dmvict: > Whatever you like. This code only to show possible usage for gas and maybe something else

Okay. But improved version may have concrete variants except gas metering and it's better to add them.

NYBACHOK: I'm unsure that we need gas in token...

PR #131, opened 2024-06-26
Validate genesis data during parsing

joneskm:
This removes the need to call validate on the genesis struct during init_genesis, which removes the need to panic.

### DIFF
--- a/x/staking/src/types/domain.rs
+++ b/x/staking/src/types/domain.rs
@@ -21,7 +21,8 @@ use gears::{
 };
 use prost::Message;
 use serde::{Deserialize, Serialize};
-use std::{cmp::Ordering, str::FromStr};
+use std::{cmp::Ordering, collections::HashSet, str::FromStr};
+use thiserror::Error;
 
 /// HistoricalInfo contains header and validator information for a given block.
 /// It is stored as part of staking module's state, which persists the `n` most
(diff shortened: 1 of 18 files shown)


dmvict (approved):

x/staking/src/types/domain.rs
```diff
@@ -500,3 +501,82 @@ impl From<Validator> for ValidatorRaw {
...
 
 impl Protobuf<ValidatorRaw> for Validator {}
+
+/// [`Validators`] is a collection of [`Validator`] with some guarantees:
+/// - the collection cannot have duplicated validators by public key
+/// - no validator can be bonded and jailed at the same time // TODO: should this be a property of the validator itself?
+/// - no bonded/unbonded validator can have zero delegator shares // TODO: should this be a property of the validator itself?
+#[derive(Debug, Clone, Deserialize, Serialize, Default)]
+#[serde(try_from = "Vec<Validator>")]
+pub struct Validators(Vec<Validator>);
```
NYBACHOK, line 511: Small comment. We need to rename `SendCoins` to `Coins`.

NYBACHOK (approved):

x/staking/src/types/domain.rs
joneskm: Yes, agreed. We have two types for coin collections, we should give them compatible names.

PR #132, opened 2024-06-26
Extend bank xmod && edit range

NYBACHOK:
Remove TODO's in bank xmod && edit range to accept generic

### DIFF
--- a/x/bank/src/keeper.rs
+++ b/x/bank/src/keeper.rs
@@ -1,3 +1,4 @@
+use crate::types::iter::balances::BalanceIterator;
 use crate::types::query::{
     QueryAllBalancesRequest, QueryAllBalancesResponse, QueryBalanceRequest, QueryBalanceResponse,
     QueryDenomsMetadataResponse,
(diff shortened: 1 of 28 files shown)


x/bank/src/keeper.rs
```diff
@@ -20,19 +21,31 @@ use gears::types::base::send::SendCoins;
...
 use std::{collections::HashMap, str::FromStr};
 
 const SUPPLY_KEY: [u8; 1] = [0];
 const ADDRESS_BALANCES_STORE_PREFIX: [u8; 1] = [2];
 const DENOM_METADATA_PREFIX: [u8; 1] = [1];
 
+pub(crate) fn account_key(addr: &AccAddress) -> Vec<u8> {
+    [
+        ADDRESS_BALANCES_STORE_PREFIX.as_slice(),
+        &[addr.len()],
```
dmvict, line 43: I saw the code where this length is converted to a byte array. Please, check creation of the key.

NYBACHOK: Maybe @joneskm knows location? I'm following cosmos way due to possible difference in hash

x/bank/src/keeper.rs
```diff
@@ -80,40 +93,122 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, AK: AuthKeeper<SK, M>, M: Module> Ban
...
+        deposit: &SendCoins,
+    ) -> Result<(), AppError> {
+        let module_acc_addr = module.get_address();
+
+        let account = self
+            .auth_keeper
+            .get_account(ctx, &module_acc_addr)?
+            .unwrap(); // TODO:
+
+        match account.has_permissions("burner") {
```
joneskm, line 109: We should try to move away from using Strings for permissions.

x/bank/src/keeper.rs
```diff
@@ -485,7 +585,28 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, AK: AuthKeeper<SK, M>, M: Module>
...
+    ) -> Result<Option<Coin>, GasStoreErrors> {
+        let store = ctx.kv_store(&self.store_key);
+        let supply_store = store.prefix_store(SUPPLY_KEY);
+
+        let amount_bytes = supply_store.get(denom.as_ref().as_bytes())?;
+
+        match amount_bytes {
+            Some(bytes) => Ok(Some(Coin {
+                denom: denom.clone(),
+                amount: Uint256::from_str(&String::from_utf8_lossy(&bytes))
```
joneskm, line 604: Might be better to not use the lossy conversion here. If it's not valid utf8 the there's a bug or corruption.

x/bank/src/keeper.rs
NYBACHOK: True, but I don't want to resolve this it in this PR

x/bank/src/keeper.rs
NYBACHOK: Keypoint here is not for long, which gives slightly better code

PR #133, opened 2024-06-26
Handle proposal submission

NYBACHOK:
Handle proposal submission without connecting it to gaia, so hash won't get affected


x/gov/src/lib.rs
```diff
@@ -7,3 +13,13 @@ pub mod params;
 pub mod query;
 pub mod submission;
 pub mod types;
+
+pub trait ProposalHandler<PSK: ParamsSubspaceKey, P> {
+    fn handle<CTX: InfallibleContextMut<DB, SK>, DB: Database, SK: StoreKey>(
+        &self,
+        proposal: &P,
+        ctx: &mut CTX,
+    ) -> anyhow::Result<()>;
```
dmvict, line 22: It's not an issue. Seems like it is better to have concrete error type here.

x/gov/src/submission/handler/handler_impl.rs
```diff
@@ -0,0 +1,30 @@
...
+
+impl<PSK: ParamsSubspaceKey, T: SubmissionHandler<PSK, ParamChange<PSK>>>
+    SubmissionCheckHandler<PSK, ParamChange<PSK>> for T
+{
+    fn submission_check<PK: ParamsKeeper<PSK>>(proposal: &ParamChange<PSK>) -> bool {
+        <PK::Param as ParamsSerialize>::keys()
+            .iter()
+            .map(|this| this.as_bytes())
+            .collect::<HashSet<_>>()
+            .contains(proposal.key.as_slice())
```
dmvict, line 20: What is the type of `proposal.key`? Do we need these conversions for the check?

x/gov/src/submission/handler/handler_impl.rs
```diff
@@ -0,0 +1,30 @@
...
+    }
+}
+
+impl<PSK: ParamsSubspaceKey, T: SubmissionHandler<PSK, TextProposal>>
+    SubmissionCheckHandler<PSK, TextProposal> for T
+{
+    fn submission_check<PK: ParamsKeeper<PSK>>(_proposal: &TextProposal) -> bool {
+        true
```
dmvict, line 28: Why it is always true?

joneskm (approved):

PR #134, opened 2024-06-26
feat:  add module slashing, implement `init_genesis` method

dmvict:


x/slashing/src/keys.rs
```diff
@@ -0,0 +1,40 @@
...
+
+pub(crate) fn validator_missed_block_bit_array_key(addr: ConsAddress, index: i64) -> Vec<u8> {
+    let mut key = VALIDATOR_MISSED_BLOCK_BIT_ARRAY_KEY_PREFIX.to_vec();
+
+    let addr_bytes = Vec::from(addr);
+    let index_bytes = index.to_le_bytes();
+    let postfix = must_length_prefixed(&addr_bytes);
+
+    key.extend_from_slice(&postfix);
+    key.extend_from_slice(&index_bytes);
```
NYBACHOK, line 28: I found `[].concat()` cleaner

x/slashing/src/keeper.rs
```diff
@@ -0,0 +1,130 @@
...
+            .staking_keeper
+            .validators_iter(ctx)
+            .unwrap_gas()
+            .map(|validator| {
+                let validator = validator.unwrap_gas();
+                validator.cons_pub_key().clone()
+            })
+            .collect();
+        for pub_key in pub_keys {
+            self.add_pub_key(ctx, &pub_key);
```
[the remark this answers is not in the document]
dmvict: You're right. I had used used this approach but failed. It's borrow checker issue. I cannot have immutable and mutable borrow of context. So, I've used this two step logic.

x/slashing/src/keys.rs
dmvict: I'll update the keys creation.

x/slashing/src/keys.rs
```diff
@@ -0,0 +1,40 @@
...
+    key
+}
+
+fn must_length_prefixed(bytes: &[u8]) -> Vec<u8> {
```
joneskm, line 32: Can this accept a `ConsAddress`? Then you don't need the length checks.

x/slashing/src/types/missed_block.rs
```diff
@@ -0,0 +1,21 @@
+use gears::types::address::ConsAddress;
+use serde::{Deserialize, Serialize};
+
+/// MissedBlock contains height and missed status as boolean.
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct MissedBlock {
+    /// index is the height at which the block was missed.
+    pub index: i64,
```
joneskm, line 8: Could this be a `u32`?

x/slashing/src/types/signing_info.rs
```diff
@@ -0,0 +1,30 @@
+use gears::{tendermint::types::time::Timestamp, types::address::ConsAddress};
+use serde::{Deserialize, Serialize};
+
+/// ValidatorSigningInfo defines a validator's signing info for monitoring their
+/// liveness activity.
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct ValidatorSigningInfo {
+    pub address: ConsAddress,
+    /// Height at which validator was first a candidate OR was unjailed
+    pub start_height: i64,
```
joneskm, line 10: Same as above, could it be a `u32`? Same for `index_offset` and `missed_block_counter`.

dmvict: We can change these fields but I suggest that is better to keep original `i64` type. 

I suggest it because these structures are stored into a store during genesis.
For example, it is a call of method [`set_validator_signing_info`](https://github.com/rumos-io/gears/pull/134/files#diff-23123a007b59eddb57ccb8b58afcf300de915aabae27b715e39399f1e0aa6672R72). It stores full struct `ValidatorSigningInfo`.

x/slashing/src/keys.rs
dmvict: updated

x/slashing/src/keys.rs
dmvict: done

x/slashing/src/keys.rs
```diff
@@ -0,0 +1,27 @@
...
+
+pub(crate) fn validator_missed_block_bit_array_key(addr: ConsAddress, index: i64) -> Vec<u8> {
+    let key = VALIDATOR_MISSED_BLOCK_BIT_ARRAY_KEY_PREFIX.to_vec();
+    let index_bytes = index.to_le_bytes().to_vec();
+    let postfix = must_length_prefixed(addr);
+    [key, postfix, index_bytes].concat()
+}
+
+fn must_length_prefixed(addr: ConsAddress) -> Vec<u8> {
+    let addr_bytes = Vec::from(addr);
```
joneskm, line 24: `ConsAddress` has a len() method which returns a u8. That way you don't need the byte conversion.

dmvict: done

x/slashing/src/keeper.rs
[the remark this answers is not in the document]
NYBACHOK: Oh. Okay. But you still may do `.into_iter().for_each()`? If you want

x/slashing/src/types/signing_info.rs
NYBACHOK: I think that we have "silent agreement" to resolve such issues with [dto's](https://en.wikipedia.org/wiki/Data_transfer_object)

NYBACHOK (approved):

PR #137, opened 2024-06-28
chore: add method `begin_block` 

dmvict:


x/slashing/src/keeper.rs
```diff
@@ -87,6 +103,211 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, SSK: SlashingStakingKeeper<SK, M>, M:
...
+
+        // fetch signing info
+        let mut sign_info = self
+            .get_validator_signing_info(ctx, &cons_addr)
+            .expect("Expected signing info for validator but it is not found");
+
+        // this is a relative index, so it counts blocks the validator *should* have signed
+        // will use the 0-value default signing info if not present, except for start height
+        let params = self.slashing_params_keeper.get(ctx);
+        let index = sign_info.index_offset % params.signed_blocks_window as u32;
```
joneskm, line 128: Is it possible to make 'signed_blocks_window` a u32?

x/slashing/src/keeper.rs
```diff
@@ -87,6 +103,211 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, SSK: SlashingStakingKeeper<SK, M>, M:
...
+                .validator_by_cons_addr(ctx, &cons_addr)
+                .unwrap_gas()
+                .map(|v| v.is_jailed())
+                .unwrap_or_default();
+
+            if validator_is_jailed {
+                // Downtime confirmed: slash and jail the validator
+                // We need to retrieve the stake distribution which signed the block, so we subtract ValidatorUpdateDelay from the evidence height,
+                // and subtract an additional 1 since this is the LastCommit.
+                // Note that this *can* result in a negative "distributionHeight" up to -ValidatorUpdateDelay-1,
```
joneskm, line 211: Our `distribution_height` is a `u32` so this isn't true.

x/slashing/src/keeper.rs
```diff
@@ -87,6 +103,211 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, SSK: SlashingStakingKeeper<SK, M>, M:
...
+                // i.e. at the end of the pre-genesis block (none) = at the beginning of the genesis block.
+                // That's fine since this is just used to filter unbonding delegations & redelegations.
+
+                let distribution_height = height - VALIDATOR_UPDATE_DELAY as u32 - 1;
```
joneskm, line 215: We need to be careful this doesn't overflow.

x/slashing/src/params.rs
```diff
@@ -32,6 +32,19 @@ pub struct SlashingParams {
...
 }
 
+impl SlashingParams {
+    pub fn min_signed_per_window_u32(&self) -> u32 {
+        // NOTE: RoundInt64 will never panic as minSignedPerWindow is less than 1.
+        self.min_signed_per_window
+            .checked_mul(Decimal256::from_atomics(self.signed_blocks_window as u64, 0).unwrap())
+            .unwrap()
+            .to_string()
+            .parse::<f64>()
```
joneskm, line 42: Do they use floating point types in cosmos?

gaia-rs/tests/staking.rs
```diff
@@ -42,6 +42,7 @@ fn run_tx_local(
     home: PathBuf,
     command: GaiaTxCommands,
 ) -> anyhow::Result<Response> {
+    // a comment
```
NYBACHOK, line 45: a comment

gears/src/x/keepers/staking.rs
```diff
@@ -15,6 +15,18 @@ use crate::{
...
+/// Delay, in blocks, between when validator updates are returned to the
+/// consensus-engine and when they are applied. For example, if
+/// ValidatorUpdateDelay is set to X, and if a validator set update is
+/// returned with new validators at the end of block 10, then the new
+/// validators are expected to sign blocks beginning at block 11+X.
+///
+/// This value is constant as this should not change without a hard fork.
+/// For Tendermint this should be set to 1 block, for more details see:
+/// https://tendermint.com/docs/spec/abci/apps.html#endblock
+// TODO: may it be u32?
```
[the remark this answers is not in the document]
NYBACHOK: Why no? This value probably needs to be added to block height, which `u32`

x/slashing/src/keeper.rs
```diff
@@ -87,6 +103,211 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, SSK: SlashingStakingKeeper<SK, M>, M:
         self.slashing_params_keeper.set(ctx, genesis.params);
     }
 
+    pub fn handle_validator_signature<DB: Database>(
+        &self,
+        ctx: &mut BlockContext<'_, DB, SK>,
+        addr_bytes: prost::bytes::Bytes,
```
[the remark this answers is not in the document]
NYBACHOK: I don't see all picture, but if you need then add `From<AccAdress> for ConsAdress` than using raw bytes

x/slashing/src/keeper.rs
```diff
@@ -87,6 +103,211 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, SSK: SlashingStakingKeeper<SK, M>, M:
...
+        power: u32,
+        signed: bool,
+    ) {
+        let height = ctx.height();
+
+        // fetch the validator public key
+        let cons_addr = ConsAddress::try_from(Vec::from(addr_bytes)).unwrap();
+        self.get_pub_key(ctx, &cons_addr)
+            .expect("validator consensus address not found");
```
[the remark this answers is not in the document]
NYBACHOK: This method used in `block_end` so @dmvict  omited it

x/slashing/src/keeper.rs
```diff
@@ -87,6 +103,211 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, SSK: SlashingStakingKeeper<SK, M>, M:
...
+                sign_info.jailed_until = Timestamp {
+                    seconds: jailed_until.timestamp(),
+                    nanos: jailed_until.timestamp_subsec_nanos() as i32,
+                };
+                // We need to reset the counter & array so that the validator won't be immediately slashed for downtime upon rebonding.
+                sign_info.missed_blocks_counter = 0;
+                sign_info.index_offset = 0;
+                self.clear_validator_missed_block_bit_array(ctx, &cons_addr);
+
+                // TODO: how do we log?
```
NYBACHOK, line 270: We don't log... in most cases

x/slashing/src/keeper.rs
```diff
@@ -87,6 +103,211 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, SSK: SlashingStakingKeeper<SK, M>, M:
...
+                        &cons_addr,
+                        distribution_height,
+                        power,
+                        params.slash_fraction_downtime,
+                    )
+                    .unwrap_gas();
+                self.staking_keeper.jail(ctx, &cons_addr).unwrap_gas();
+
+                let time = ctx.get_time();
+                // TODO: consider to move the DateTime type and work with timestamps into Gears
```
NYBACHOK, line 255: Open issue for it

gears/src/x/keepers/staking.rs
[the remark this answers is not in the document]
dmvict: I'll change.

x/slashing/src/abci_handler.rs
```diff
@@ -26,4 +27,24 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, SSK: SlashingStakingKeeper<SK, M>, M:
...
+        &self,
+        _ctx: &mut BlockContext<'_, DB, SK>,
+        request: RequestBeginBlock,
+    ) {
+        // Iterate over all the validators which *should* have signed this block
+        // store whether or not they have actually signed it and slash/unbond any
+        // which have missed too many blocks in a row (downtime slashing)
+        if let Some(vote_info) = request.last_commit_info {
+            for vote in vote_info.votes {
+                // TODO: seems like tendermint type has optional value in order of using prost
```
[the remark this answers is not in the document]
dmvict: Our domain type [inherits tendermint](https://github.com/rumos-io/gears/blob/main/tendermint/src/types/proto/info.rs#L35). Should we change the type?

x/slashing/src/keeper.rs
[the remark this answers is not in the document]
dmvict: I've used type from our domain type [`Validator` type](https://github.com/rumos-io/gears/blob/main/tendermint/src/types/proto/validator.rs#L39). I suggest that we can update both info and validator type. I mean non optional validator field with type containing an address (ConsAddr for now).

x/slashing/src/keeper.rs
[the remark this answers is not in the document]
dmvict: Yes, the method is used by `begin_block` method and should be infallible. I think it is better to save this behavior because the method is used only by `begin_block`.

x/slashing/src/keeper.rs
dmvict: I think that params struct should have types similar to cosmos and it is better to save type.

 I can change it to `u32` and then I have to change serialization/deserialization of params. 

What do you think about it?

x/slashing/src/keeper.rs
dmvict: I'll fix

x/slashing/src/params.rs
dmvict: They don't. There was no easy way to get i64 from the Decimal256 and I used this conversion. You're right, I have to implement better solution here.

gaia-rs/tests/staking.rs
dmvict: I'll remove

x/slashing/src/keeper.rs
dmvict: :ok_hand:

x/slashing/src/abci_handler.rs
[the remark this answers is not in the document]
NYBACHOK: If it only for prost then yes. That's why this domain types exist, but I hadn't enough time to edit them.

gaia-rs/tests/staking.rs
NYBACHOK: F for `a comment`

x/slashing/src/keeper.rs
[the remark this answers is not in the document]
joneskm: I don't think you have enough context to panic at this point. The method accepts an arbitrary validator which may or may not exist, you can't tell from inside the method whether it is expected to exist.

gears/src/x/keepers/staking.rs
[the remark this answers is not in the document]
dmvict: Fixed

x/slashing/src/abci_handler.rs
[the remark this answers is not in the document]
dmvict: Updated.

x/slashing/src/keeper.rs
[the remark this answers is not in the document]
dmvict: I investigated this part and see that we can't change type because it is part of `tendermint` crate. The `AccAddress` and the other address types are placed in `gears`. It's not a good idea to create circular dependency. 

We can add some wrapper in `gears` . I think it is an extra work and I think it is better to save this type and this conversion. 

Alternative solution. We can add a type `Address` in tendermint that will wrap the bytes and validate them like `BaseAddress` in `gears`. After that we can add `impl From<Address> for ConsAddress in `gears`.

@joneskm, @NYBACHOK What do you think?

x/slashing/src/keeper.rs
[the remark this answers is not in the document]
dmvict: Updated.

x/slashing/src/keeper.rs
dmvict: done

x/slashing/src/keeper.rs
dmvict: done

x/slashing/src/params.rs
dmvict: Updated

gaia-rs/tests/staking.rs
dmvict: done

x/slashing/src/keeper.rs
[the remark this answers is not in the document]
NYBACHOK: Are there issue with moving `BaseAdress<>` into our tendermint crate?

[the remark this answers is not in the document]
joneskm: This has been resolved by introducing a new `address` crate.

x/slashing/src/abci_handler.rs
```diff
@@ -26,4 +27,31 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, SSK: SlashingStakingKeeper<SK, M>, M:
...
+            for vote in vote_info.votes {
+                self.keeper
+                    .handle_validator_signature(
+                        ctx,
+                        vote.validator.address.into(),
+                        vote.validator.power as u32,
+                        vote.signed_last_block,
+                    )
+                    .expect(
+                        "method `handle_validator_signature` is called from infallible method.
```
joneskm, line 51: We try to follow the advice given [here](https://doc.rust-lang.org/std/result/enum.Result.html) on how to use expect: "We recommend that expect messages are used to describe the reason you expect the `Result` should be `Ok`". I understand that not all projects use `expect` this way but I think we should. In which case you'll have to explicitly panic here - assuming that's how cosmos handles this situation.

PR #138, opened 2024-06-28
Validation of values in proposal process

NYBACHOK:


gears/src/application/keepers/params.rs
```diff
@@ -16,6 +18,17 @@ pub trait ParamsKeeper<PSK: ParamsSubspaceKey> {
...
+    fn check_key<SL: AsRef<[u8]>>(key: SL) -> bool {
+        <Self::Param as ParamsSerialize>::keys()
+            .iter()
+            .map(|this| this.as_bytes())
+            .collect::<HashSet<_>>()
+            .contains(key.as_ref())
+    }
+
+    #[cfg(all(feature = "governance"))]
+    fn validate(key: impl AsRef<[u8]>, value: impl AsRef<[u8]>) -> bool;
```
dmvict, line 30: How it can resolve the type behind the bytes? Is it required?

NYBACHOK: Developer resolves

PR #139, opened 2024-06-29
Cli & Query for governance xmod

NYBACHOK:


gears/src/rest/pagination.rs
```diff
@@ -25,6 +25,23 @@ pub fn parse_pagination(pagination: Pagination) -> (u32, u8) {
...
+        core_types::query::request::PageRequest {
+            key: _,
+            offset,
+            limit,
+            count_total: _,
+            reverse: _,
+        }: core_types::query::request::PageRequest,
+    ) -> Self {
+        Self {
+            offset: Some(offset as u32),
```
joneskm, line 39: These casts can wrap around, which is probably not the desired behaviour.

gears/src/ext.rs
```diff
@@ -0,0 +1,33 @@
...
+/// - `T`: The input `Option`'s value type
+/// - `U`: The outputs `Option`'s value type
+/// - `E`: The possible error during the mapping
+pub trait FallibleMapExt<T, U, E> {
+    /// Try to apply a fallible map function to the option
+    fn try_map<F>(self, f: F) -> Result<Option<U>, E>
+    where
+        F: FnOnce(T) -> Result<U, E>;
+}
+
```
joneskm, line 22: Can you achieve something similar using collect?

NYBACHOK: It's possible.  Collect `Vec<Result<T, ()>>` iterate over elements and push them to new collection

gears/src/rest/pagination.rs
NYBACHOK: I'll try to change it there

NYBACHOK: Okay. I tried to make it `u64`, but we depend on tendermint in [this](https://github.com/NYBACHOK/gears/blob/c97a2a209acf0b192859781d2c6deaa767544451/gears/src/rest/handlers.rs#L69) code, so I have chosen to set max values if couldn't narrow number

x/gov/src/params.rs
```diff
@@ -21,14 +26,17 @@ const DEFAULT_PERIOD: Duration = Duration::from_secs(172800); // 2 days
...
+    pub min_deposit: SendCoins,
     pub max_deposit_period: Duration, // ?
 }
 
 impl Default for DepositParams {
     fn default() -> Self {
         Self {
-            min_deposit: vec![Coin::from_str("10000000stake").expect("default is valid")],
+            min_deposit: SendCoins::new(vec![
+                Coin::from_str("10000000stake").expect("default is valid")
```
dmvict, line 37: Is it from sdk? As I remember we use `uatom`

NYBACHOK: [It's](https://github.com/cosmos/cosmos-sdk/blob/d3f09c222243bb3da3464969f0366330dcb977a8/x/gov/types/params.go#L51-L55). Think I should change to `uatom`?

dmvict (approved):

x/gov/src/params.rs
dmvict: I don't know. Maybe we can use this denom.

joneskm: For now just make it "uatom". We might need a better solution in future.

NYBACHOK: Done

PR #140, opened 2024-07-01
Abci testing

joneskm:


gaia-rs/tests/abci/main.rs
```diff
@@ -0,0 +1,131 @@
...
+        db,
+        GaiaParamsStoreKey::BaseApp,
+        GaiaABCIHandler::new(config),
+        node_options,
+    );
+    let chain_id = ChainId::default();
+
+    let mnemonic = "race draft rival universe maid cheese steel logic crowd fork comic easy truth drift tomorrow eye buddy head time cash swing swift midnight borrow";
+    let mnemonic = bip32::Mnemonic::new(mnemonic, bip32::Language::English).unwrap();
+    let key_pair = KeyPair::from_mnemonic(&mnemonic);
```
dmvict, line 58: As I remember we have similar code in `keypair`. And if I'm correct, we have to reuse it (maybe update before)

gears/src/baseapp/abci.rs
```diff
@@ -206,7 +206,7 @@ impl<DB: Database, PSK: ParamsSubspaceKey, H: ABCIHandler, AI: ApplicationInfo>
     fn commit(&self) -> ResponseCommit {
         info!("Got commit request");
 
-        let height = self.block_height();
+        let height = self.get_block_header().unwrap().height;
```
NYBACHOK, line 209: unwrap?

gaia-rs/tests/abci/main.rs
joneskm: Yes [here](https://github.com/rumos-io/gears/blob/0045ac03ad7a5dce396164b3d7f16c5012d5919b/keyring/src/key/pair/mod.rs#L68)

gears/src/baseapp/abci.rs
```diff
@@ -220,7 +220,7 @@ impl<DB: Database, PSK: ParamsSubspaceKey, H: ABCIHandler, AI: ApplicationInfo>
 
         ResponseCommit {
             data: hash.to_vec().into(),
-            retain_height: (height - 1), // TODO: check if this is correct - why -1?
+            retain_height: 0, // this is the height above which tendermint will retain all blocks // TODO: make this configurable as in Cosmos
```
NYBACHOK, line 223: I have a feeling that we have more TODO that open issues. Maybe its worth to open one for this "issue"?

gears/src/baseapp/mod.rs
```diff
@@ -75,13 +73,6 @@ impl<DB: Database, PSK: ParamsSubspaceKey, H: ABCIHandler, AI: ApplicationInfo>
             .map(|e| e.max_gas)
             .unwrap_or_default();
 
-        block_height_set(multi_store.head_version());
```
NYBACHOK, line 78: Are you sure about this change? Where block_heigh should be set?

gears/src/baseapp/abci.rs
```diff
@@ -234,7 +234,7 @@ impl<DB: Database, PSK: ParamsSubspaceKey, H: ABCIHandler, AI: ApplicationInfo>
     fn begin_block(&self, request: RequestBeginBlock) -> ResponseBeginBlock {
         info!("Got begin block request");
 
-        self.block_height_increment();
```
NYBACHOK, line 237: ? Probably you moved it to `end_block`

gears/src/baseapp/abci.rs
joneskm: It's effectively a TODO. It would be good to find a way of making the header not be an `Option`. One way might be to initialize it to an "empty" header.

gears/src/baseapp/abci.rs
joneskm: You're right. I'll open an issue.

gears/src/baseapp/mod.rs
joneskm: We no longer track the height since we can get it from the stored header or the request. This is inline with how Cosmos does it. We do now have a method for getting the last commit height of the store.

dmvict (approved):

Issue #91, opened 2024-05-21
[TODO]: Use `HashSet` for `SendCoins` and `MinGasPrices` with custom `Ord` impl for `Coins`


NYBACHOK:

NYBACHOK:
Implemented in https://github.com/rumos-io/gears/pull/155, but reverted due to possible contract issue with cosmos.

PR #155, opened 2024-07-10
Refactor coins types

NYBACHOK:
Resolves: https://github.com/rumos-io/gears/issues/91


dmvict (approved):

gears/src/types/base/coin/unsigned.rs
```diff
@@ -14,53 +14,64 @@ mod inner {
...
-pub struct Coin {
+#[serde(try_from = "inner::Coin", into = "inner::Coin")]
+pub struct UnsignedCoin {
     pub denom: Denom,
     pub amount: Uint256,
 }
 
-impl TryFrom<inner::Coin> for Coin {
-    type Error = CoinsError;
+impl Coin<Uint256> for UnsignedCoin {
```
joneskm, line 23: This makes me think that `Coin` amount type should not be a generic argument but an associated type

NYBACHOK: Makes sense

NYBACHOK: Done.

PR #160, opened 2024-07-11
Add min-gas-fees to config

NYBACHOK:


gears/src/types/base/min_gas.rs
```diff
@@ -1,53 +1,70 @@
...
+            amount: Decimal256::zero(),
+        }])
     }
+}
+
+impl MinGasPrices {
+    pub fn new(coins: impl IntoIterator<Item = DecimalCoin>) -> Result<Self, CoinsError> {
+        let coins = coins.into_iter().collect::<Vec<_>>();
 
-    // Checks that the SendCoins are sorted, have positive amount, with a valid and unique
```
dmvict, line 24: It is better to add a part of the comment as doc

joneskm (commented):
The run command should fail if the min gas is not set on the CLI and there's no min gas in the config. I think you assume a default value in such cases.

PR #180, opened 2024-07-13
Pagination for queries

NYBACHOK:


gears/src/rest/pagination.rs
```diff
@@ -1,21 +1,39 @@
...
-
-//#[derive(FromForm, Debug)]
-#[derive(Deserialize, serde::Serialize, Debug, Clone, Eq, PartialEq)]
+#[derive(serde::Deserialize, serde::Serialize, Debug, Clone, Eq, PartialEq)]
 pub struct Pagination {
-    offset: Option<u32>,
+    /// offset is a numeric offset that can be used when key is unavailable.
+    /// It is less efficient than using key. Only one of offset or key should
+    /// be set.
+    pub offset: u32,
```
joneskm, line 8: Doesn't this change the API?

NYBACHOK: How? Behavior same, except now you write `Option<Pagination>`

joneskm: I think a query like this would have worked previously: `?offset=10` i.e. the offset would be set to ten. Now I assume the offset would be set to default?

NYBACHOK: Hm. I see what you mean. You right and for rest it would be better to have optional fields, or use `serde(default)`

NYBACHOK: After a minute I think that for rest `Pagination` with `Option` fields would be okay.

x/bank/src/keeper.rs
```diff
@@ -420,62 +420,45 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, AK: AuthKeeper<SK, M>, M: Module>
...
         let prefix = create_denom_balance_prefix(addr);
         let account_store = bank_store.prefix_store(prefix);
+        let total = account_store.clone().into_range(..).count();
 
         let mut balances = vec![];
-        for rcoin in account_store.into_range(..) {
+
+        let iterator = match pagination {
+            Some(pagination) => {
+                TwoIterators::One(account_store.into_range(..).paginate(pagination))
```
joneskm, line 438: Can we hide this from the user i.e. we handle pagination=None in gears?

x/bank/src/keeper.rs
```diff
@@ -486,7 +463,18 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, AK: AuthKeeper<SK, M>, M: Module>
                     .unwrap_or_corrupt();
                 UnsignedCoin { denom, amount }
             })
-            .collect()
+            .filter(|this| !this.amount.is_zero());
+
+        let total = supply_store.clone().count();
+
+        let mut store: Vec<_> = match pagination {
+            Some(pagination) => supply_store.paginate(pagination).collect(),
```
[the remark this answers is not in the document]
NYBACHOK: Yes

dmvict (approved):

PR #186, opened 2024-07-14
Mark method for governance as unsafe

NYBACHOK:


joneskm:
I don't think we want any unsafe code in the crate especially since this isn't really unsfae.

NYBACHOK:
Okay. I'll think about any other way

PR #188, opened 2024-07-14
Remove possible overflow for gas metering

NYBACHOK:


gears/src/types/auth/gas.rs
```diff
@@ -51,6 +51,20 @@ impl Gas {
...
 
+    // TODO: write a test for this
+    pub fn checked_sub(self, rhs: Self) -> Option<Self> {
+        let self_inner: u64 = self.0.into();
+        let rhs_inner: u64 = rhs.0.into();
+
+        let result = self_inner.checked_sub(rhs_inner)?;
+
+        if result > u63::MAX.into() {
+            None
```
joneskm, line 62: This is unreachable!

NYBACHOK: What do you mean? I'm adding 2 `u64` so overflow could happen

joneskm: I meant specifically line 62

PR #191, opened 2024-07-14
Remove todo for auth message enum

NYBACHOK:


joneskm (commented):
Maybe we should delete the entire `x/auth/src/message.rs` file?

NYBACHOK:
> Maybe we should delete the entire `x/auth/src/message.rs` file?

@joneskm well. We could. Seems this enum didn't use in any place

PR #192, opened 2024-07-14
Panic if try set empty key

NYBACHOK:


joneskm (commented):
Do you know why it's bad to set an empty key? I would like to understand before merging.

NYBACHOK:
> Do you know why it's bad to set an empty key? I would like to understand before merging.

Not really. Replication of cosmos logic

PR #198, opened 2024-07-16
Paginate queries by key

NYBACHOK:


gears/src/types/pagination/request.rs
```diff
@@ -1,64 +1,82 @@
...
             limit,
             count_total: _,
             reverse: _,
         }: core_types::query::request::PageRequest,
     ) -> Self {
         Self {
-            offset: offset.try_into().unwrap_or(u32::MAX),
+            kind: match key.is_empty() {
+                true => PaginationKind::Key {
+                    key: key.try_into().expect("we checked it for empty"),
```
joneskm, line 61: Are true/false the wrong way round here?

NYBACHOK: fixed

PR #204, opened 2024-07-19
feat: add support of multiple messages into transaction handler

dmvict:


joneskm (approved):

gears/src/types/tx/errors.rs
```diff
@@ -1,5 +1,9 @@
 use crate::types::auth::info::AuthError;
 
+#[derive(Debug, thiserror::Error)]
+#[error("transaction applies non empty set of messages")]
+pub struct MessagesError;
```
NYBACHOK, line 5:
```suggestion
pub struct EmptyMessagesError;
```

gears/src/types/tx/mod.rs
```diff
@@ -26,6 +29,53 @@ pub trait TxMessage:
     fn type_url(&self) -> &'static str;
 }
 
+/// Utility type that guarantees correctness of transaction messages set
+pub struct Messages<T: TxMessage> {
+    messages: Vec<T>,
```
NYBACHOK, line 34: Not sure, that you want, but there `Vec1` type

x/gov/src/client/tx_handler.rs
```diff
@@ -35,23 +40,26 @@ impl TxHandler for GovClientHandler {
                 proposal_id,
                 depositor: from_address,
                 amount,
-            })),
+            })
+            .into()),
```
NYBACHOK, line 44: May you collect this into variable and do `into` in one place?

gears/src/types/tx/errors.rs
dmvict: It's the good one.

gears/src/types/tx/mod.rs
dmvict: We suggested to save this struct.

x/gov/src/client/tx_handler.rs
dmvict: Thanks, it's better solution.

NYBACHOK (approved):

PR #207, opened 2024-07-22
feat: add cli to `distribution` module

dmvict:


gears/src/types/base/coins/decimal.rs
```diff
@@ -20,6 +20,10 @@ impl DecimalCoins {
         }
 
         for coin in other {
+            // Note:
+            // operator '>' means that all values are greater or equal
```
joneskm, line 24: I find these statements a bit confusing. Are they referring to the `is_all_gte` method?

gears/src/types/base/coins/decimal.rs
```diff
@@ -79,7 +83,12 @@ impl DecimalCoins {
 
     pub fn checked_sub(&self, other: &DecimalCoins) -> Result<Self, CoinsError> {
         if self.is_all_gte(other.inner()) {
-            let coins = self.checked_calculate_iterate(other.inner(), Decimal256::checked_sub)?;
+            let coins: Vec<DecimalCoin> = self
```
joneskm, line 86: Looks like we need documentation and tests around these methods to define the behaviour. Especially since it looks like the behaviour is changed as part of this PR.

gears/src/types/decimal256.rs
```diff
@@ -4,6 +4,8 @@ pub use cosmwasm_std::Decimal256;
 use cosmwasm_std::{StdError, Uint256};
 
 pub const ONE_DEC: Decimal256 = Decimal256::one();
+// 1/10^18
+pub const SMALLEST_DEC: Decimal256 = Decimal256::new(Uint256::from_u128(1u128));
```
joneskm, line 8: Why not use https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.Decimal256.html#associatedconstant.MIN

x/distribution/src/keeper/validator.rs
```diff
@@ -0,0 +1,170 @@
...
+        period: u64,
+    ) -> Result<(), AppError> {
+        let mut historical = self
+            .validator_historical_rewards(ctx, validator_address, period)?
+            .ok_or(AppError::Custom(
+                "historical rewards are not found".to_string(),
+            ))?;
+        if historical.reference_count > 2 {
+            // TODO: sdk behaviour, seems to be correct
+            panic!("reference count should never exceed 2")
```
joneskm, line 31: This is a bit worrying. Do you expect it not to happen? Can we capture this in a type?

x/distribution/src/abci_handler.rs
```diff
@@ -1,40 +1,69 @@
...
         keepers::{
-            auth::AuthKeeper, bank::StakingBankKeeper as BankKeeper, staking::SlashingStakingKeeper,
+            auth::AuthKeeper, bank::StakingBankKeeper as BankKeeper,
+            staking::DistributionStakingKeeper,
         },
         module::Module,
     },
 };
 
+#[derive(Clone)]
```
NYBACHOK, line 28: We should forbid creation of non `Debug` struct and enums...

x/distribution/src/abci_handler.rs
```diff
@@ -44,6 +73,128 @@ impl<
...
+
+                Ok(self
+                    .keeper
+                    .query_delegator_validators(ctx, req)
+                    .encode_vec()
+                    .into())
+            }
+            "/cosmos.distribution.v1beta1.Query/CommunityPool" => {
+                let req = QueryCommunityPoolRequest::decode(query.data)
+                    .map_err(|e| CoreError::DecodeProtobuf(e.to_string()))?;
```
NYBACHOK, line 148: Note: if you merge upstream line `.map_err(|e| CoreError::DecodeProtobuf(e.to_string()))?;` could be omitted.

gears/src/types/base/coins/decimal.rs
dmvict: Yes, it refers to method `is_all_gte` that means all values of another coins are greater or equal 
```rs
            if coin.amount > self.amount_of(&coin.denom) {
                return false;
            }
```

This code means that method returns false for exactly greater values of other coins. 
For example;
```rs 
let a = vec![2,3,4];
let b = vec![1, 3, 2]; 
assert!(a.is_all_gte(b));
assert!(!a.is_all_gt(b));

let a = vec![2,3,4];
let b = vec![1, 3, 5]; 
assert!(!a.is_all_gte(b));
assert!(!a.is_all_gt(b));
```

gears/src/types/base/coins/decimal.rs
dmvict: The logic is covered by tests and added some documentation.

gears/src/types/decimal256.rs
dmvict: The `Decimal256::MIN` is 0. The sdk has another value that is written here. 

```rs
    let min = Decimal256::new(Uint256::from_u128(1u128));
    let min_2 = Decimal256::MIN;
    dbg!(&min, &min_2);
```
Output:
```
[src/main.rs:6:5] &min = Decimal256(0.000000000000000001)
[src/main.rs:6:5] &min_2 = Decimal256(0)
```

joneskm: Ah I think I understand. `SMALLEST_DEC` is the minimum positive value.

x/distribution/src/errors.rs
```diff
@@ -17,3 +27,49 @@ pub enum TokenAllocationError {
     #[error("{0}")]
     BankSend(#[from] BankKeeperError),
 }
+
+#[derive(Debug, Clone, thiserror::Error)]
+pub enum DistributionError {
```
NYBACHOK, line 32: Um... are you sure that it should be single error?

x/distribution/src/keeper/mod.rs
```diff
@@ -87,11 +93,12 @@ impl<
         ctx: &mut InitContext<'_, DB, SK>,
         genesis: GenesisState,
     ) -> anyhow::Result<()> {
-        self.set_fee_pool(ctx, &genesis.fee_pool);
+        self.set_fee_pool(ctx, &genesis.fee_pool).unwrap_gas();
```
NYBACHOK, line 96: There is huge number of `unwrap_gas` in PR

x/distribution/src/keeper/query.rs
```diff
@@ -0,0 +1,181 @@
...
+            ctx,
+            &self.store_key,
+            &validator_address,
+            starting_height,
+            ending_height,
+        )
+        .maybe_paginate(pagination.map(Pagination::from));
+
+        let mut events = vec![];
+        for res in slash_events_iterator {
```
NYBACHOK, line 69: Maybe we need extension to filter out such items?

x/distribution/src/keeper/validator.rs
```diff
@@ -0,0 +1,149 @@
...
+        let mut historical = if let Some(rewards) =
+            self.validator_historical_rewards(ctx, validator_operator_addr, period)?
+        {
+            rewards
+        } else {
+            return Err(DistributionError::ValidatorHistoricalRewardsNotFound(
+                validator_operator_addr.clone(),
+            ));
+        };
+        // TODO: panics in sdk
```
NYBACHOK, line 134: Reminder: omit usage of panic, as they do in sdk. For us, panic is programmer fault.

x/distribution/src/types/query.rs
NYBACHOK, line 1: It looks like xmods that you work doesn't have raw variant from ibc-proto. What do you think, should macro generating this be prioritized?

NYBACHOK (approved):

x/distribution/src/errors.rs
dmvict: 90% of errors I added following single method that withdraw rewards.  I've added only single error from the other method.

x/distribution/src/keeper/mod.rs
dmvict: I've used it only from `InitContext`, `BlockContext` and `QueryContext` methods. The unwraps are legal here.

x/distribution/src/keeper/validator.rs
dmvict: I do not panic. But it may be a valid way to stop chain.

x/distribution/src/types/query.rs
dmvict: I've checked the ibc and didn't find raw variants for the structs. It's not necessary but should improve the coding.

x/distribution/src/keeper/validator.rs
NYBACHOK: I mean, you left comment about panic, so I mentioned it

x/distribution/src/types/query.rs
NYBACHOK: @joneskm should it be my next task?

x/distribution/src/errors.rs
NYBACHOK: Wow

PR #210, opened 2024-07-25
feat: extend `auth` commands

dmvict:


x/auth/src/client/grpc.rs
```diff
@@ -46,7 +54,9 @@ where
         let req = AuthNodeQueryRequest::Account(request.into_inner().try_into()?);
         let response = self.app.typed_query(req)?;
         let response: AuthNodeQueryResponse = response.try_into()?;
-        let AuthNodeQueryResponse::Account(response) = response;
+        let AuthNodeQueryResponse::Account(response) = response else {
+            return Err(Status::unknown("invalid response enum variant"));
```
joneskm, line 58: We usually return an `internal` error here e.g. https://github.com/rumos-io/gears/blob/abe901d3886b32481eeeea7e7c9796bb9e2805e6/x/bank/src/client/grpc.rs#L46-L48

x/auth/src/client/grpc.rs
```diff
@@ -46,7 +54,9 @@ where
         let req = AuthNodeQueryRequest::Account(request.into_inner().try_into()?);
         let response = self.app.typed_query(req)?;
         let response: AuthNodeQueryResponse = response.try_into()?;
-        let AuthNodeQueryResponse::Account(response) = response;
+        let AuthNodeQueryResponse::Account(response) = response else {
+            return Err(Status::internal("invalid response enum variant"));
```
joneskm, line 58: I think it's better to be consistent and say "An internal error occurred while querying the application state."

core-types/src/query/response/auth.rs
```diff
@@ -0,0 +1,3 @@
+pub use ibc_proto::cosmos::auth::v1beta1::QueryAccountResponse;
```
NYBACHOK, line 1: Are you sure that we should re-export this structures?

dmvict: The highlighted struct has been imported. I've added new structs and they are used in grpc handlers. I may skip reexporting but I have to import them directly.

x/auth/src/client/grpc.rs
dmvict: done

PR #212, opened 2024-07-25
Derive macro for raw prost structures

NYBACHOK:


macros/protobuf-derive/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+struct OptionalOrRepeated {
+    optional: Flag,
+    repeated: Flag,
+}
+
+impl OptionalOrRepeated {
+    fn not_both(self) -> darling::Result<Self> {
+        if self.optional.is_present() && self.repeated.is_present() {
+            Err(
+                darling::Error::custom("Cannot set `optional` and `repeated`")
```
dmvict, line 33: I'm not sure about this case. It seems like can exist field `Option<Vec<msg>>`. 

For now we can save this restriction.

macros/protobuf-derive/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    fn check_for_option(path: &syn::Path) -> Option<&syn::PathSegment> {
+        let idents_of_path = path.segments.iter().fold(String::new(), |mut acc, v| {
+            acc.push_str(&v.ident.to_string());
+            acc.push(':');
+            acc
+        });
+        vec!["Option:", "std:option:Option:", "core:option:Option:"]
```
dmvict, line 102: Technically we can have custom `Option`, we can omit the case. 
But here you compares concrete values and missed `option::Option:`. I suggest that the check can be improved. Something like `idents_of_path.contains("::Option") || idents_of_path.starts_with("Option")`

macros/protobuf-derive/src/raw.rs
```diff
@@ -0,0 +1,138 @@
...
+            };
+
+            Ok(gen.into())
+        }
+        _ => Err(syn::Error::new(
+            proc_macro2::Span::call_site(),
+            "Protobuf can be derived only for `struct`",
+        )),
+    }
+}
```
dmvict, line 138: It's a good idea to add tests for the macro.

macros/protobuf-derive/src/lib.rs
NYBACHOK: If we have custom `Option` it's porblem 'cause I use `::std::option::Option` in macro code, and it would result in error

macros/protobuf-derive/src/lib.rs
NYBACHOK: Technically, it is possible even now.  `#[raw(raw = Vec::<T>, optional)]`

macros/protobuf-derive/src/raw.rs
NYBACHOK: Would be great if you know which crate to use for it

NYBACHOK:
There is other issue which same for all our macroses. We can't use them in gears 'cause we use absolute path's

macros/protobuf-derive/src/raw.rs
dmvict: `trybuild` is used for this purpose.

macros/protobuf-derive/src/lib.rs
dmvict: At least you have to update condition here and for the vec. You've missed one case as I described above.

NYBACHOK: Okay

NYBACHOK: Added other variants

PR #221, opened 2024-08-01
Change protobuf trait we use

NYBACHOK:


gears/src/commands/client/tx.rs
```diff
@@ -35,10 +34,7 @@ pub struct ClientTxContext {
 }
 
 impl ClientTxContext {
-    pub fn query<
-        Response: Protobuf<Raw> + std::convert::TryFrom<Raw>,
```
NYBACHOK, line 39: I removed `Protobuf` bound not only 'cause it's makes no sense for us, but rust don't happy with trait bounds for Error.

tendermint/Cargo.toml
```diff
@@ -29,6 +29,7 @@ rand = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 prost = { workspace = true }
+ibc-proto = { workspace = true } # We only need Protobuf trait...
```
NYBACHOK, line 32: Maybe it makes sense to create our copy of `Protobuf` trait?

tendermint/src/types/time/timestamp.rs
```diff
@@ -336,7 +335,8 @@ impl From<Timestamp> for inner::Timestamp {
     }
 }
 
-impl Protobuf<Timestamp> for Timestamp {}
+impl tendermint_proto::Protobuf<Timestamp> for Timestamp {}
```
dmvict, line 338: Do we need both versions?

x/auth/src/keeper.rs
```diff
@@ -83,7 +82,7 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, M: Module> AuthKeeper<SK, M> for Keep
         let mut auth_store = ctx.kv_store_mut(&self.store_key);
         let key = create_auth_store_key(acct.get_address().to_owned());
 
-        auth_store.set(key, acct.encode_vec().expect(IBC_ENCODE_UNWRAP))?; // TODO:IBC
+        auth_store.set(key, acct.encode_vec())?;
```
dmvict, line 85: Why here we propagate error?

NYBACHOK: Check code again. It was before me

dmvict (approved):

tendermint/src/types/time/timestamp.rs
joneskm: I think we might need the tendermint variant inside our tendermint crate

PR #222, opened 2024-08-01
Joneskm/querying

joneskm:


address/src/lib.rs
```diff
@@ -72,6 +72,14 @@ impl<const PREFIX: u8> BaseAddress<PREFIX> {
             .expect("MAX_ADDR_LEN is a u8 so this can't fail")
     }
 
+    /// Returns the address bytes prefixed with the length of the address.
+    pub fn prefix_len_bytes(&self) -> Vec<u8> {
+        let len = self.len();
+        let mut prefixed = vec![len];
+        prefixed.extend_from_slice(&self.0);
```
NYBACHOK, line 79: What about `[&[prefixed], &self.0].concat()`?

docs/tutorials/staking.md
```diff
@@ -0,0 +1,27 @@
...
+  --broadcast-mode block \
+  -y
+```
+
+```
+gaiad query staking validators
+```
+
+```
+gaiad export
```
NYBACHOK, line 26: export?

gears/src/crypto/public.rs
```diff
@@ -22,12 +30,77 @@ impl PublicKey {
...
+            },
+            PublicKey::Ed25519(key) => Any {
+                type_url: "/cosmos.crypto.ed25519.PubKey".to_string(),
+                value: key.encode_vec(),
+            },
+        }
+    }
+}
+
+use tendermint::types::proto::crypto::PublicKey as TendermintPublicKey;
```
NYBACHOK, line 84: Interesting that for this you didn't follow my way with `inner` mod.

x/staking/Cargo.toml
```diff
@@ -18,3 +18,7 @@ ibc-proto = { workspace = true }
 tracing = { workspace = true }
 tonic = { workspace = true }
 tower-layer = { workspace = true }
+serde-aux = { workspace = true }
+
+[dev-dependencies]
+hex = { workspace = true }
```
NYBACHOK, line 24: I think that for new deps crate  `data-enoding`  is more preferable.

NYBACHOK: https://docs.rs/data-encoding/latest/data_encoding/

address/src/lib.rs
joneskm: To make it work I have to do `[prefix, self.0.clone()].concat()`. This could require two allocations of `self.0`. How about:
```
       let len = self.len();
        let mut prefixed = Vec::with_capacity(len as usize + 1);
        prefixed.push(len);
        prefixed.extend_from_slice(&self.0);
        prefixed
        ```

File: docs/tutorials/staking.md (hunk 2 above)
Comment (joneskm) [reply docs/tutorials/staking.md:26]: I will remove this

File: gears/src/crypto/public.rs (hunk 3 above)
Comment (joneskm) [reply gears/src/crypto/public.rs:84]: I think the use case is a bit different. We usually use inner because the user doesn't need to know about that type. In this case users are aware of the type - we're giving them From and Into methods for the type.

File: address/src/lib.rs (hunk 1 above)
Comment (NYBACHOK) [reply address/src/lib.rs:79]: Check [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=aa7d5b668d1dd009048c53022a7fd036) playground.

Comment (joneskm) [reply address/src/lib.rs:79]: Went with:
```
let len = self.len();
[&[len], self.0.as_slice()].concat()
```
Suggested by @NYBACHOK

x/staking/Cargo.toml
joneskm: Changed to use `data-encoding`.

NYBACHOK (approved):

gears/src/crypto/public.rs
NYBACHOK: Not sure that this is 100% correct, 'cause you don't re-export this type. And `inner` for imports is not requirements. Just a way which easiest for me and I noted that you and @dmvict  follow it.

dmvict (approved):

PR #227, opened 2024-08-06
feat: add module `evidence`

dmvict:
The module has a few main features/disadvantages: 
- used `Any` in genesis and generic for transforming it to concrete type 
- runtime validation of genesis structs 
- router that applies generics related to context `DB, SK` and propagates the generics on `Keeper` and `ABCIHandler` structs. We can't skip it because we don't know these types. And here I've added struct `Handler` with quite ugly api. I was trying to hide the concrete context types and do not extend number of generics in the struct 
- used `Rc` for handler routines. We may replace it to `Arc` if it is required.


x/evidence/src/keeper.rs
```diff
@@ -0,0 +1,127 @@
...
+            self.router = Some(router);
+            Ok(())
+        }
+    }
+
+    /// genesis initializes the evidence module's state from a provided genesis
+    /// state.
+    pub fn init_genesis(&self, ctx: &mut InitContext<'_, DB, SK>, genesis: GenesisState) {
+        if let Err(e) = genesis.validate::<E>() {
+            panic!("failed to validate evidence genesis state: {e}");
```
NYBACHOK, line 88: What do you think? Is it worth to return errors from keeper and panic only in abci handler or it more annoying?

x/evidence/src/types/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+mod query;
+mod router;
+mod tx;
+pub use router::*;
+
+//
+
+pub trait Evidence: Message {
+    fn route(&self) -> String;
+    fn r#type(&self) -> String;
```
NYBACHOK, line 13: kind?

x/evidence/src/types/router.rs
```diff
@@ -0,0 +1,229 @@
...
+    Block(Rc<HandlerFn<BlockContext<'a, DB, SK>, E>>),
+}
+
+impl<'a, DB, SK, E> Handler<'a, DB, SK, E>
+where
+    E: Evidence,
+    SK: StoreKey,
+    DB: Database,
+{
+    pub fn new_init(handler_fn: HandlerFn<InitContext<'a, DB, SK>, E>) -> Self {
```
NYBACHOK, line 50: Hmmm. This all structure is... okay. What about implementing `From`?

NYBACHOK:
This code is not common for us, but hard to say without all context why you chose it. Maybe you could describe which issues you have with generics and etc?

x/evidence/src/types/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+    fn height(&self) -> i64;
+}
+
+pub trait EvidenceHandler<E: Evidence>
+where
+    <E as std::convert::TryFrom<Any>>::Error: std::fmt::Debug,
+{
+    type Error;
+
+    fn handle<CTX: TransactionalContext<DB, SK>, DB: Database, SK: StoreKey>(
```
joneskm, line 35: Do you prefer to have two traits or could the `handle` be part of the `Evidence` trait?

dmvict: Yeah. It's good idea. I've merged the traits.

x/evidence/src/types/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+where
+    <Self as std::convert::TryFrom<Any>>::Error: std::fmt::Debug,
+{
+    // TODO: uncomment or remove
+    // fn route(&self) -> String;
+    // Original method is named `type`, replaced as inner interface
+    fn kind(&self) -> String;
+    fn string(&self) -> String;
+    fn hash(&self) -> Hash;
+    fn validate_basic(&self) -> anyhow::Result<()>;
```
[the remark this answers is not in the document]
dmvict: In this case I saved the function. We can't force a user to implement validation of his evidence in implementation of `TryFrom<Any>`. 
We can remove the member if we force users to use our evidence type (actually Tendermint evidence type), but then we can skip all generics in the module.

x/evidence/src/genesis.rs
```diff
@@ -0,0 +1,25 @@
+use gears::core::any::google::Any;
+use serde::{Deserialize, Serialize};
+
+use crate::Evidence;
+
+/// GenesisState defines the evidence module's genesis state.
+#[derive(Debug, Clone, Default, PartialEq, Deserialize, Serialize)]
+pub struct GenesisState {
+    /// evidence defines all the evidence at genesis.
```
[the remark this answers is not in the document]
dmvict: Added and removed validation.

x/evidence/src/types/router.rs
dmvict: non actual

x/evidence/src/keeper.rs
dmvict: Moved

x/evidence/src/types/mod.rs
dmvict: Replaced

x/evidence/src/types/mod.rs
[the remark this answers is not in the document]
NYBACHOK: > but then we can skip all generics in the module.
Isn't it good?

x/evidence/src/genesis.rs
```diff
@@ -0,0 +1,49 @@
...
+where
+    <E as std::convert::TryFrom<Any>>::Error: std::fmt::Debug,
+{
+    /// evidence defines all the evidence at genesis.
+    pub evidence: Evidences<E>,
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, Default)]
+#[serde(try_from = "Vec<Any>")]
+pub struct Evidences<E: Evidence>(Vec<E>)
```
joneskm, line 17: Would it make sense to have the `#[serde(try_from = "Vec<Any>")]` on the Gensis struct and avoid having the `Evidences` struct altogether?

x/evidence/src/types/mod.rs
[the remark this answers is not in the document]
joneskm: I would have thought that most users would be familiar with the idea of doing validation in a `TryFrom` implementation.

joneskm (approved):

NYBACHOK (approved):

PR #230, opened 2024-08-08
ABCI tests

NYBACHOK:


gears/src/types/tx/mod.rs
```diff
@@ -27,6 +27,35 @@ pub trait TxMessage:
...
+
+impl From<NullTxMsg> for Any {
+    fn from(_: NullTxMsg) -> Self {
+        unreachable!()
+    }
+}
+
+impl TxMessage for NullTxMsg {
+    fn get_signers(&self) -> Vec<&AccAddress> {
+        unreachable!()
```
dmvict, line 51: I understand that the struct and impls are intended for testing purpose. Does ante handler runs checks for this struct? If yes, then you have to add impl of some methods.

gears/src/baseapp/params.rs
```diff
@@ -293,4 +302,91 @@ mod tests {
...
+
+    #[test]
+    fn set_read_works() {
+        let keeper = BaseAppParamsKeeper {
+            params_subspace_key: SubspaceKey::Params,
+        };
+
+        let mut multi_store = ApplicationMultiBank::<_, SubspaceKey>::new(MemDB::new());
+
+        let mut ctx = InitContext::new(
```
dmvict, line 379: It seems like you may reuse your method in `utils`. I might be wrong because I don't know about configuration features here.

x/auth/src/params.rs
```diff
@@ -197,3 +197,83 @@ impl<PSK: ParamsSubspaceKey> ParamsKeeper<PSK> for AuthParamsKeeper<PSK> {
...
+    fn app_hash() {
+        let keeper = AuthParamsKeeper {
+            params_subspace_key: SubspaceKey::Auth,
+        };
+
+        let mut multi_store = ApplicationMultiBank::<_, SubspaceKey>::new(MemDB::new());
+
+        let before_hash = multi_store.head_commit_hash();
+
+        let mut ctx = InitContext::new(
```
dmvict, line 223: Here you've used `gears` with `utils` feature. Please, reuse or update and reuse helper.

x/auth/tests/abci.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    let (mut node, _) = init_node(opt);
+
+    let app_hash = node.step(vec![], Timestamp::UNIX_EPOCH);
+    assert_eq!(
+        data_encoding::HEXLOWER.encode(app_hash),
+        "8d3663f81a98bec58a6d3a9f39c38469438bfecd8257dd335c8c047b933b08ad"
+    );
+
+    for _ in 0..100 {
```
dmvict, line 39: Please, add a method to node to skip steps. Something like: 
`node.skip_steps(100)`

x/auth/tests/abci.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+#[derive(strum::EnumIter, Debug, PartialEq, Eq, Hash, Clone)]
+pub enum SpaceKey {
+    Auth,
+    Params,
+}
+
+#[derive(strum::EnumIter, Debug, PartialEq, Eq, Hash, Clone)]
+pub enum SubspaceKey {
```
dmvict, line 85: I'm not sure that we need similar enums for each test file

gears/src/types/tx/mod.rs
NYBACHOK: This isn't possible to execute at all. Empty enum couldn't exist

gears/src/baseapp/params.rs
NYBACHOK: This isn't possible. `utils` under feature while this compiles always

x/auth/tests/abci.rs
NYBACHOK: Ok

x/auth/tests/abci.rs
NYBACHOK: I'm unsure how to do this. Maybe new crate with all stuff from `utils` + these enum with all possible xmods?

dmvict: At least we can use one extended enum for tests inside mod. What do you think? 
But it is better to use single enum for all tests because it allows to reduce boilerplate.

x/auth/tests/abci.rs
NYBACHOK: Done

x/auth/tests/abci.rs
NYBACHOK: I want to avoid any connection between xmods, but this creates such boilerplate. Main the issue is that we should create sort of mocks for our traits. Enum is the smallest issue that may be fixed with one large variant for test purposes

x/auth/tests/abci.rs
joneskm: The begin/end block will still need to run. So timestamps for every block will be needed

dmvict (approved):

PR #240, opened 2024-08-16
feat: add `bank` `params` query, improve a rest handler

dmvict:


x/bank/src/client/rest.rs
```diff
@@ -25,15 +25,14 @@ pub async fn supply<
     QRes: QueryResponse,
     App: NodeQueryHandler<QReq, QRes>,
 >(
-    _pagination: Query<Pagination>,
+    pagination: Query<Pagination>,
     State(rest_state): State<RestState<QReq, QRes, App>>,
 ) -> Result<Json<QRes>, HTTPError> {
     let req = BankNodeQueryRequest::TotalSupply(QueryTotalSupplyRequest {
-        pagination: None, //pagination.0.map(PaginationRequest::from), //TODO fix this
```
NYBACHOK, line 32: ?

x/bank/src/params.rs
```diff
@@ -33,18 +47,17 @@ impl ParamsSerialize for BankParams {
...
-
-        hash_map
+        vec![
+            (
+                KEY_DEFAULT_SEND_ENABLED,
+                self.default_send_enabled.to_string().into_bytes(),
+            ),
+            (
+                KEY_SEND_ENABLED,
+                serde_json::to_vec(&self.send_enabled)
```
joneskm, line 57: We should keep one of the TODOs because it's still true, maybe with some rewording:
```TODO: if params are missing in the cosmos SDK (e.g the send_enabled field is missing from the genesis json file) then they are set to "null" i.e. [110, 117, 108, 108] when stored. This is different to the Gears behaviour which would fail to parse if e.g. the send enabled field is missing
```

dmvict: Added

joneskm (approved):

NYBACHOK (approved):

PR #242, opened 2024-08-20
Fix bug with order

NYBACHOK:


kv_store/src/bank/kv/application.rs
```diff
@@ -141,17 +142,26 @@ impl<DB: Database> ApplicationKVBank<DB> {
...
 
         let mut persistent = self.persistent.write().expect(POISONED_LOCK);
 
-        cache
+        let cache = insert
             .into_iter()
             .filter(|(key, _)| !delete.contains(key))
-            .for_each(|(key, value)| persistent.set(key, value));
+            .map(|(key, value)| (key, Some(value)))
+            .collect::<Vec<_>>()
```
joneskm, line 153: Is there a reason to collect to vec then go back to an iter?

kv_store/src/bank/kv/transaction.rs
```diff
@@ -53,7 +53,6 @@ impl<DB: Database> TransactionKVBank<DB> {
     pub fn upgrade_cache(&mut self) {
         let (set_values, delete) = self.tx.take();
         for (key, value) in set_values {
-            self.block.delete.remove(&key);
```
joneskm, line 56: Is that part of the same fix?

kv_store/src/bank/kv/application.rs
NYBACHOK: Borrowing of `delete`.

kv_store/src/bank/kv/transaction.rs
NYBACHOK: This is code what should have been removed long ago. It duplicates logic from cache

kv_store/src/bank/kv/application.rs
NYBACHOK: W8. It shouldn't make any sense 'cause I remove equal keys on set or delete

PR #245, opened 2024-08-22
feat: add rest endpoints to transfer tokens via frontend

dmvict:


gears/src/rest/handlers.rs
```diff
@@ -80,6 +85,82 @@ pub async fn txs<M: TxMessage>(
     Ok(Json(res))
 }
 
+pub async fn tx<M: TxMessage>(
```
NYBACHOK, line 88: Now we're using async. Unexpected

NYBACHOK (approved):

gears/src/signing/std_sign_doc.rs
```diff
@@ -0,0 +1,114 @@
+use serde::{Deserialize, Serialize};
+use serde_json::{Map, Value};
+
+// gears::core::base::coin::Coin has wrong order of fields
+// It is better to create a struct with correct order than
```
joneskm, line 5: I don't think there would be a problem with changing the order in gears Coins [here](https://github.com/rumos-io/gears/blob/72baade2fd17cd9a91c7fc83f74a662544352f0a/gears/src/types/base/coin/unsigned.rs#L18)

gears/src/x/ante.rs
```diff
@@ -450,6 +451,60 @@ impl<
...
                     }
                     .encode_to_vec(),
+                    SignMode::LegacyAminoJson => {
+                        let mut msgs = vec![];
+                        for msg in tx.tx.get_msgs() {
+                            // TODO: don't have proper value getter, maybe extend trait
+                            let mut value: serde_json::Map<String, serde_json::Value> =
+                                serde_json::from_slice(
+                                    &serde_json::to_vec(&msg)
+                                        .map_err(|_| AnteError::LegacyAminoJson)?,
```
joneskm, line 461: We could do with a better solution here... maybe we have a `AminoSerde` trait, kind of like our `ValueRenderer` trait

gears/src/x/ante.rs
```diff
@@ -450,6 +451,60 @@ impl<
...
+                            })
+                        }
+                        let amount = {
+                            let fee = tx
+                                .tx
+                                .auth_info
+                                .fee
+                                .amount
+                                .clone()
+                                .ok_or(AnteError::MissingFee)?;
```
joneskm, line 480: There is no requirement for a Tx to contain a fee

gears/src/x/ante.rs
```diff
@@ -450,6 +451,60 @@ impl<
...
+                                    amount: coin.amount.to_string(),
+                                    denom: coin.denom.to_string(),
+                                })
+                                .collect()
+                        };
+                        let doc = std_sign_doc::StdSignDoc {
+                            account_number: acct.get_account_number().to_string(),
+                            chain_id: ctx.chain_id().to_string(),
+                            fee: std_sign_doc::StdFee {
+                                amount,
```
joneskm, line 493: Is it possible to implement `From<GearsFee> for StdFee`

gears/src/signing/std_sign_doc.rs
dmvict: The structure requires strings instead of domain types. May I add custom serialization/deserialization for struct like`#[serde(try_from`?

gears/src/x/ante.rs
dmvict: Your comment is related to my `TODO`:                            
```rs
   // TODO: don't have proper value getter, maybe extend trait
```

You're right, it's better to add some proper solution for serializing domain tx types into amino types.

gears/src/x/ante.rs
dmvict: The fee is required by `StdSignDoc`. At least I understand it by this way. I'll check/update code.

gears/src/x/ante.rs
dmvict: I'll add.

gears/src/signing/std_sign_doc.rs
joneskm: Why does the structure require Strings? Doesn't it just need to serialize to JSON?

gears/src/x/ante.rs
```diff
@@ -454,56 +455,28 @@ impl<
...
-                            // TODO: don't have proper value getter, maybe extend trait
-                            let mut value: serde_json::Map<String, serde_json::Value> =
-                                serde_json::from_slice(
-                                    &serde_json::to_vec(&msg)
-                                        .map_err(|_| AnteError::LegacyAminoJson)?,
-                                )
-                                .map_err(|_| AnteError::LegacyAminoJson)?;
-                            value.remove("@type");
-
+                            dbg!(msg.get_signers());
```
NYBACHOK, line 458: You forget to remove `dbg` macro

gears/src/types/response/tx.rs
```diff
@@ -193,6 +193,8 @@ impl<M: TxMessage> From<TxResponse<M>> for TxResponseRaw {
 /// Service.BroadcastTx method.
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct BroadcastTxResponse<M: TxMessage> {
+    /// tx represents the transaction that is placed inside tx_response
+    pub tx: Option<Tx<M>>,
```
NYBACHOK, line 197: Unusual, but should be fine.

joneskm (approved):

PR #246, opened 2024-08-22
Enable staking begin block

joneskm:


NYBACHOK (approved):

x/staking/src/types/historical_info.rs
```diff
@@ -36,3 +39,42 @@ impl HistoricalInfo {
         HistoricalInfo { header, validators }
     }
 }
+
+mod inner {
+    pub use gears::tendermint::types::proto::header::HistoricalInfo;
+}
+
+impl From<HistoricalInfo> for inner::HistoricalInfo {
```
dmvict, line 47: It seems like the derive macro `Protobuf` should work here. Not an issue, it's a suggestion.

dmvict (approved):

x/staking/src/types/historical_info.rs
joneskm: I've updated to use the `Protobuf` macro.

PR #251, opened 2024-09-02
Update tx macro & usage on structs [NOT READY]

NYBACHOK:


dmvict (approved):

joneskm (approved):

gaia-rs/src/message.rs
```diff
@@ -8,14 +8,14 @@ use gears::{
 };
 use serde::Serialize;
 
-#[derive(Debug, Clone, RoutingMessage, Serialize)]
+#[derive(Debug, Clone, AppMessage, Serialize)]
 #[serde(untagged)]
 pub enum Message {
-    #[gears(url = "/cosmos.bank.v1beta1")]
+    #[msg(url(string = "/cosmos.bank.v1beta1"))]
```
dmvict, line 14: It looks strange. Why did you use so complex syntax?

dmvict (approved):

gaia-rs/src/message.rs
NYBACHOK: To allow string and TypePath. Gaia part tries to parse any with prefix, while lower type by matching type(usually)

PR #252, opened 2024-09-03
feat: add endpoints for main page, update modules codebase

dmvict:


gears/src/rest/handlers.rs
```diff
@@ -51,6 +52,50 @@ pub async fn node_info(
     Ok(Json(node_info))
 }
 
+pub async fn validatorsets_latest(
```
NYBACHOK, line 55: Should this `fn` be part of gears?

x/staking/src/keeper/mod.rs
```diff
@@ -145,6 +146,8 @@ impl<
                 BondStatus::Unbonding | BondStatus::Unbonded => {
                     not_bonded_tokens += validator.tokens;
                 }
+                // TODO: maybe move panics to abci handler
```
NYBACHOK, line 149: Yes. You need to move them out of keeper layer.

x/staking/src/keeper/query.rs
```diff
@@ -63,11 +101,80 @@ impl<
...
+        let store = store.prefix_store(DELEGATION_KEY);
+        let key = query.delegator_addr.prefix_len_bytes();
+        let (p_result, iterator) = store
+            .into_range(..)
+            .maybe_paginate(query.pagination.map(gears::ext::Pagination::from));
+
+        let delegation_responses = iterator
+            .filter_map(|(k, bytes)| {
+                if k.starts_with(&key) {
+                    Delegation::decode_vec(&bytes).ok()
```
NYBACHOK, line 124: Is it okay to ignore invalid keys?

gears/src/rest/pagination.rs
```diff
@@ -9,12 +9,20 @@ pub struct Pagination {
     /// offset is a numeric offset that can be used when key is unavailable.
     /// It is less efficient than using key. Only one of offset or key should
     /// be set.
+    #[serde(alias = "pagination.offset")]
     offset: Option<u32>,
     /// limit is the total number of results to be returned in the result page.
     /// If left empty it will default to a value to be set by each app.
+    #[serde(alias = "pagination.limit")]
```
NYBACHOK, line 16: I'm unsure if `alias` attribute overrides default name. If so, would it be better to add the original name too?

dmvict: It doesn't override the original name. It allows to deserialize a struct with such name into destination struct. I see no reasons for additional alias because it is a default value.

x/staking/src/keeper/query.rs
dmvict: I'm not sure that it is better than an error. I'll update.

gears/src/rest/handlers.rs
dmvict: Actually not. It's the call to `tendermint` service and it should be only an endpoint that connects app with tendermint. The scheme with `tendermint` looks like:

tendermint asks app for validators to create validatorset -> frontend asks app to ask tendermint for validatorset

Btw, each app with staking module may ask for validators by `validators` query.

joneskm (approved):

NYBACHOK (approved):
@joneskm, I discussed PR with @dmvict and have no issues with this PR, but wait confirmation for merge

PR #256, opened 2024-09-04
Implement auth grpc

NYBACHOK:


x/auth/src/client/grpc.rs
```diff
@@ -90,35 +91,57 @@ where
         &self,
         _request: Request<QueryModuleAccountsRequest>,
     ) -> Result<Response<QueryModuleAccountsResponse>, Status> {
-        unimplemented!() //TODO: implement
+        // TODO: Discuss implementation of this. Easiest way is to implement them with strum enum iterator
```
NYBACHOK, line 94: I want to hear your opinion about this. [This](https://github.com/cosmos/cosmos-sdk/blob/7799bba7bc889288607576aa11655b5fc31a2da9/x/auth/keeper/keeper.go#L186-L207) is cosmos impl

joneskm: This endpoint was added after v45 of the SDK so we can leave it unimplemented for now.

PR #267, opened 2024-09-12
Fix bug in tree

joneskm:
The leftmost key in the subtree was incorrectly set as the root key.


trees/src/iavl/tree.rs
```diff
@@ -179,6 +179,40 @@ impl Node {
...
+            Node::Leaf(leaf) => return leaf.key.clone(),
+            Node::Inner(inner) => {
+                let mut left_node = match &inner.left_node {
+                    Some(inner_left_node) => inner_left_node,
+                    None => &node_db
+                        .get_node(&inner.left_hash)
+                        .expect("node db should contain all nodes"),
+                };
+
+                let mut cached_node;
```
NYBACHOK, line 194: You may use `break` to return value from `loop`

NYBACHOK (approved):

trees/src/iavl/tree.rs
joneskm: Changed.

PR #268, opened 2024-09-12
Extend test beyond unbondng time + fixes

joneskm:


x/staking/src/keeper/unbonding.rs
```diff
@@ -443,7 +441,7 @@ impl<
     ) -> Result<(), GasStoreErrors> {
         let store = TransactionalContext::kv_store_mut(ctx, &self.store_key);
         let mut store = store.prefix_store_mut(VALIDATOR_QUEUE_KEY);
-        store.delete(&validator_queue_key(&end_time, end_height))?;
+        store.delete(&validator_queue_key(&end_time, end_height))?; //TODO: should we NOT be using the prefix store here - since the validator_queue_key adds the prefix?
```
NYBACHOK, line 444: What about app hash? If it correct then no, but in other case we should omit prefix store,

x/staking/src/keys.rs
```diff
@@ -54,11 +54,9 @@ pub(super) fn parse_validator_queue_key(key: &[u8]) -> anyhow::Result<(Timestamp
     Ok((time, height))
 }
 
+// This is the key for use in the unbonding queue sub store (UNBONDING_QUEUE_KEY prefix)
 pub(super) fn unbonding_delegation_time_key(time: &Timestamp) -> Vec<u8> {
-    let tbz = time.format_bytes_rounded();
-    let mut bz = UNBONDING_QUEUE_KEY.to_vec();
-    bz.extend(tbz.iter());
-    bz
+    time.format_bytes_rounded()
```
NYBACHOK, line 59: Maybe it better to remove such method completely?

NYBACHOK (approved):

x/staking/src/keeper/unbonding.rs
joneskm: I need to check whether we have a test which takes this path. If we do then I'll remove the TODO, if we don't then I'll add a test.

x/staking/src/keys.rs
joneskm: I've kept it in for now...

PR #284, opened 2024-09-24
fix: Fix redelegations

dmvict:


x/staking/src/keeper/mod.rs
```diff
@@ -416,11 +416,11 @@ impl<
         let mut amt_from_not_bonded_to_bonded = Uint256::zero();
 
         let mut last = self.last_validators_by_addr(ctx);
-        let validators_map = self.validators_power_store_vals_map(ctx)?;
+        let validators_map = self.validators_power_store_vals_vec(ctx)?;
 
         let mut updates = vec![];
 
-        for (_k, val_addr) in validators_map.iter().take(max_validators as usize) {
+        for val_addr in validators_map.iter().take(max_validators as usize).rev() {
```
NYBACHOK, line 423: Shouldn't it `rev` and only after that limit?

dmvict: Good catch. I've just append the call but it's not a correct change.

NYBACHOK (approved):

PR #285, opened 2024-09-24
feat: Improve handler for `txs` endpoint, allow to pass multiple events 

dmvict:


gears/src/rest/tendermint_events_handler.rs
```diff
@@ -0,0 +1,33 @@
...
+    events_str: &'a str,
+}
+
+impl<'a> StrEventsHandler<'a> {
+    pub fn new(events_str: &'a str) -> Self {
+        let events_str = events_str.trim_start_matches('\'').trim_end_matches('\'');
+        Self { events_str }
+    }
+
+    pub fn try_parse_tendermint_events_vec(&self) -> anyhow::Result<Vec<String>> {
```
NYBACHOK, line 14: Would it make sense to return `Vec<(String : attribute, String : value)>` instead of string concatenation?

gears/src/rest/tendermint_events_handler.rs
```diff
@@ -0,0 +1,33 @@
...
+
+        let mut tm_events = Vec::with_capacity(self.events_str.matches('&').count() + 1);
+        for event in events {
+            if !event.contains('=') {
+                return Err(anyhow!("invalid event; event {event} should be of the format: {{eventType}}.{{eventAttribute}}={{value}}"));
+            }
+            let tokens = event
+                .split_once('=')
+                .expect("the condition is checked above");
+            if tokens.0 == "tx.height" {
```
NYBACHOK, line 25: Should every `value` which is number follow this rule?

NYBACHOK: With smth like this: https://doc.rust-lang.org/stable/std/primitive.char.html#method.is_numeric

gears/src/rest/tendermint_events_handler.rs
dmvict: We could use this solution but it's much easier to use Query parser.

gears/src/rest/tendermint_events_handler.rs
dmvict: It shouldn't follow the rule. The query has only one exception for `tx.height` event type.

NYBACHOK: I should have guessed that this is weird requirement from cosmos.

NYBACHOK (approved):

PR #286, opened 2024-09-24
chore: Update type `TxResponse`

dmvict:


NYBACHOK (approved):

gears/src/types/response/tx.rs
```diff
@@ -100,6 +102,14 @@ impl<M: TxMessage> TryFrom<TxResponseRaw> for TxResponse<M> {
             events,
         }: TxResponseRaw,
     ) -> Result<Self, Self::Error> {
+        let mut events_res = vec![];
```
joneskm, line 105: Can we get rid of the conversion on lines 76-78

dmvict: Updated.

PR #289, opened 2024-09-25
fix: Fix staking params deserialization, fix query test

dmvict:


x/staking/src/params/mod.rs
```diff
@@ -34,6 +35,32 @@ pub struct StakingParams {
...
+    historical_entries: u32,
+    bond_denom: Denom,
+}
+
+impl TryFrom<RawStakingParams> for StakingParams {
+    type Error = anyhow::Error;
+
+    fn try_from(params: RawStakingParams) -> Result<Self, Self::Error> {
+        StakingParams::new(
+            i128::from(params.unbonding_time.duration_nanoseconds())
```
joneskm, line 53: Can we make the `StakingParams::new` method's `unbonding_time` argument type be `Duration`? We might need to derive `Ord` on `Duration`?

joneskm:
Merged as part of https://github.com/rumos-io/gears/pull/290

PR #292, opened 2024-09-26
Resolve issue for end_block don't use recent changes

NYBACHOK:


kv_store/src/bank/multi/transaction.rs
```diff
@@ -43,4 +45,15 @@ impl<DB: Database, SK: StoreKey> MultiBank<DB, SK, TransactionStore<DB, SK>> {
             store.append_block_cache(other.kv_store_mut(sk))
         }
     }
+
+    pub fn take_block_cache(
```
dmvict, line 49: It seems like we do many extra clones here. Do we? Maybe it's not an issue, I just want to understand is it okay or not.

NYBACHOK: This is not a clone (except SK which is simple enum), but you're right that all this stuff needs to be optimized at some point.

PR #294, opened 2024-09-27
Polish auth xmod

NYBACHOK:


gears/src/baseapp/genesis.rs
```diff
@@ -8,6 +8,7 @@ pub struct GenesisError(pub AccAddress);
 pub trait Genesis:
     std::fmt::Debug + Default + DeserializeOwned + Serialize + Clone + Send + Sync + 'static
 {
+    // TODO: REMOVE?
```
NYBACHOK, line 11: Is there any way to rework this? As I see this method needed by 2 xmods only.

x/auth/src/keeper.rs
```diff
@@ -144,118 +139,65 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, M: Module> Keeper<SK, PSK, M> {
         }
     }
 
-    pub fn init_genesis<DB: Database>(
+    pub fn init<DB: Database, CTX: TransactionalContext<DB, SK>>(
```
joneskm, line 142: Why `TransactionalContext` vs `InitContext`?

NYBACHOK: I split single method into 2 parts and wanted to make keeper as more generic as possible

NYBACHOK: As you see, I even removed `_genesis` part from name

Makefile
```diff
@@ -8,7 +8,7 @@ run:
 	cargo run -- run --min-gas-prices 0uatom
 
 test:
-	cargo test --features=macros_test,it  --no-fail-fast 
+	RUST_BACKTRACE=1 cargo test --features=macros_test,it  --no-fail-fast
```
dmvict, line 11: I don't think that it's necessary. You may replace it by environment from the command.

dmvict (approved):

Makefile
NYBACHOK: True, but on the other hand if run test you want to see where error occurred without re-running them.

x/auth/src/keeper.rs
NYBACHOK: I can't say `yes` or `no`. But probably there wouldn't be such cases

joneskm: > ...make keeper as more generic as possible
I generally aim for the exact opposite :wink:

NYBACHOK: Okay. I revert some changes

NYBACHOK: Edited to use InitContext and removed gas

PR #295, opened 2024-09-30
Polish bank xmod

NYBACHOK:


x/bank/src/types/iter/balances.rs
```diff
@@ -18,7 +18,7 @@ impl<'a, DB: Database> BalanceIterator<'a, DB> {
     pub fn new(store: Store<'a, DB>, addr: &AccAddress) -> BalanceIterator<'a, DB> {
         let store = store.prefix_store(account_key(addr));
 
-        // TODO: WHY https://github.com/cosmos/cosmos-sdk/blob/d3f09c222243bb3da3464969f0366330dcb977a8/store/prefix/store.go#L88-L93
+        // https://github.com/cosmos/cosmos-sdk/blob/d3f09c222243bb3da3464969f0366330dcb977a8/store/prefix/store.go#L88-L93
```
joneskm, line 21: What's the URL for?

NYBACHOK: The creation of key, but this iterator is not being used.

x/bank/src/keeper/bank.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        self.sub_unlocked_coins(ctx, &module_acc_addr, deposit)?;
+
+        for coin in deposit.inner() {
+            let supply = self.supply(ctx, &coin.denom)?;
+            // we slightly different from cosmos in this. They return coin always.
+            // If no value found then new coin with zero balance, subtraction and call of set_supply which deletes coins with zero balance
+            // We omitted it but if any issue arise be aware that maybe we should delete zero coins if we store any.
+            if let Some(mut supply) = supply {
+                supply.amount.sub_assign(coin.amount);
```
joneskm, line 76: Does this panic if supply < coin.amount?

NYBACHOK: This type from `cosmwasms-std`. I think it follows standart logic for rust. panic for debug overflow for release

x/bank/src/genesis.rs
```diff
@@ -29,9 +28,6 @@ impl Default for GenesisState {
                 send_enabled: vec![],
                 default_send_enabled: true,
             },
-            //TODO: this denom metadata should not be hard coded into the bank module
-            // this has been added here for short term convenience. There should be a
```
joneskm, line 33: Also remove the commented out lines below?

NYBACHOK: Only when you or me updates init scripts to set metadata in genesis

x/bank/src/keeper/mod.rs
```diff
@@ -287,16 +92,20 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, AK: AuthKeeper<SK, M>, M: Module>
...
+        mut balances: Vec<Balance>,
+        params: BankParams,
+        denom_metadata: Vec<Metadata>,
     ) {
-        // TODO:
-        // 1. cosmos SDK sorts the balances first
-        // 2. Need to confirm that the SDK does not validate list of coins in each balance (validates order, denom etc.)
-        // 3. Need to set denom metadata
-        self.bank_params_keeper.set(ctx, genesis.params);
+        // 1. cosmos SDK sorts the balances first - Make sure that rust ordering gives same result
```
joneskm, line 99: This should still be a TODO

NYBACHOK: Okay. But I checked and only 1 is worth todo, others correct

gears/src/x/keepers/staking.rs
```diff
@@ -227,6 +228,9 @@ pub trait StakingBankKeeper<SK: StoreKey, M: Module>:
         amount: UnsignedCoins,
     ) -> Result<(), BankKeeperError>;
 
+    /// Method undelegates the unbonding coins and transfers
+    /// them from a module account to the delegator account. It will panic if the
```
joneskm, line 232: Does it actually panic?

x/bank/src/keeper/mod.rs
```diff
@@ -597,7 +353,7 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, AK: AuthKeeper<SK, M>, M: Module>
         ctx: &mut CTX,
         coin: UnsignedCoin,
     ) -> Result<(), GasStoreErrors> {
-        // TODO: need to delete coins with zero balance
+        // If there will be any issue with bank then delete coins with zero balance. See for more details: https://github.com/NYBACHOK/gears/blob/c79c50f60e40cee3fbc050002bf1ce63a72b67a3/x/bank/src/keeper/bank.rs#L70-L72
```
joneskm, line 356: There is an issue. Coins with zero balance need to be deleted.

x/bank/src/keeper/mod.rs
```diff
@@ -309,19 +118,14 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, AK: AuthKeeper<SK, M>, M: Module>
...
-            self.set_supply(
-                ctx,
-                UnsignedCoin {
-                    denom: coin.0,
-                    amount: coin.1,
-                },
-            )
-            .unwrap_gas();
+        // does the SDK sort these?
+        // No. It uses ordering from balances https://github.com/cosmos/cosmos-sdk/blob/d3f09c222243bb3da3464969f0366330dcb977a8/x/bank/keeper/genesis.go#L32-L34
```
joneskm, line 122: We use a HashMap so we can't guarantee the same order

x/bank/src/keeper/mod.rs
```diff
@@ -287,16 +92,20 @@ impl<SK: StoreKey, PSK: ParamsSubspaceKey, AK: AuthKeeper<SK, M>, M: Module>
...
-        // 1. cosmos SDK sorts the balances first
-        // 2. Need to confirm that the SDK does not validate list of coins in each balance (validates order, denom etc.)
-        // 3. Need to set denom metadata
-        self.bank_params_keeper.set(ctx, genesis.params);
+        // 1. cosmos SDK sorts the balances first - Make sure that rust ordering gives same result
+        // 2. Need to confirm that the SDK does not validate list of coins in each balance (validates order, denom etc.) - Yes it does and our Coins type did it
+        // 3. Need to set denom metadata - dedicated cmd for it
+        self.bank_params_keeper.set(ctx, params);
+
+        // Make sure that this 100% same as in cosmos(generally it should)
```
joneskm, line 104: This looks like a TODO?

x/bank/src/keeper/mod.rs
NYBACHOK: And you do this, but on the other level.

x/bank/src/keeper/mod.rs
NYBACHOK: I'm more confident in this rather than don't. So don't think this is todo

x/bank/src/keeper/mod.rs
NYBACHOK: Added TODO

x/bank/src/keeper/mod.rs
NYBACHOK: Okay. Vec is a bad idea.

NYBACHOK: Will create a new PR with dedicated structure for maybe empty coins vec

joneskm (approved):

gears/src/types/base/coin/decimal.rs
```diff
@@ -121,3 +121,9 @@ impl FromStr for DecimalCoin {
         Ok(Self { denom, amount })
     }
 }
+
+impl std::fmt::Display for DecimalCoin {
```
dmvict, line 125: It seems like you can implement it once for the struct `Coins`

x/bank/src/aux/metadata.rs
```diff
@@ -0,0 +1,93 @@
...
+            }
+            None => {
+                original_meta.insert(meta.name.clone(), meta);
+            }
+        }
+    }
+
+    if !dry {
+        *genesis
+            .pointer_mut("/app_state/bank/denom_metadata")
```
dmvict, line 77: May it be a configurable path or a const in the app level?

x/bank/src/keeper/bank.rs
```diff
@@ -0,0 +1,132 @@
...
+        deposit: &UnsignedCoins,
+    ) -> Result<(), BankKeeperError> {
+        let module_acc_addr = module.address();
+
+        let account = self
+            .auth_keeper
+            .get_account(ctx, &module_acc_addr)?
+            .ok_or(AccountNotFound::new(module_acc_addr.to_string()))?;
+
+        match account.has_permissions("burner") {
```
dmvict, line 63: Not an issue of the PR. We have to add enum for the permissions.

x/bank/src/aux/metadata.rs
NYBACHOK: Yes

gears/src/types/base/coin/decimal.rs
NYBACHOK: I may, but for now there is no need for it. It will do this in separate PR

x/bank/src/aux/metadata.rs
NYBACHOK: Done

dmvict (approved):

PR #298, opened 2024-10-01
Persist changes from ante 

NYBACHOK:


joneskm:
Merged as part of https://github.com/rumos-io/gears/pull/304

PR #313, opened 2024-10-04
feat: add config for client

dmvict:


gears/src/cli/query.rs
```diff
@@ -1,33 +1,41 @@
...
+use std::marker::PhantomData;
 use tendermint::types::proto::block::Height;
 
 /// Querying subcommands
 #[derive(Debug, Clone, ::clap::Args)]
-pub struct CliQueryCommand<C: Subcommand> {
+pub struct CliQueryCommand<T: ApplicationInfo, C: Subcommand> {
     /// <host>:<port> to Tendermint RPC interface for this chain
-    #[arg(long, global = true, action = ArgAction::Set, value_hint = ValueHint::Url, default_value_t = DEFAULT_TENDERMINT_RPC_ADDRESS.parse().expect( "const should be valid"))]
+    #[arg(long, global = true, action = ArgAction::Set, value_hint = ValueHint::Url, default_value_t = std::env::var("GEARS_NODE").map(|v| v.parse().expect("GEARS_NODE should be a valid http/https url")).unwrap_or(client_config(&T::home_dir()).node()))]
```
NYBACHOK, line 12: I'm impressed that I made cli which allows smth like that, but it would hard to read. Could you add dedicated method for it?

dmvict: Fixed. 
I've used `clap` attribute.

gears/src/cli/query.rs
```diff
@@ -1,33 +1,41 @@
...
+};
 use clap::{ArgAction, Subcommand, ValueHint};
+use std::marker::PhantomData;
 use tendermint::types::proto::block::Height;
 
 /// Querying subcommands
 #[derive(Debug, Clone, ::clap::Args)]
-pub struct CliQueryCommand<C: Subcommand> {
+pub struct CliQueryCommand<T: ApplicationInfo, C: Subcommand> {
     /// <host>:<port> to Tendermint RPC interface for this chain
```
joneskm, line 11: I think we should clarify the behaviour here e.g.
`Overrides any node in the client config. Default value is used if neither this argument nor a config value is provided`
+ Does the env variable override all of them.

dmvict: Added doc

PR #317, opened 2024-10-09
Add gov xmod to gaia-rs

joneskm:


x/gov/src/keeper.rs
```diff
@@ -132,6 +138,10 @@ impl<
         }
         self.gov_params_keeper.set(ctx, params);
 
+        self.auth_keeper
```
NYBACHOK, line 141: This should be a part of gaia? You added a new keeper only for simple creation of account.

joneskm: For now I thin it would be best to stick to how Cosmos do it i.e. inside the module.

